PR #4, opened 2016-06-24
Clingcon beta

MaxOstrowski:
Hiho Rolando,

my first pull request.
I added a Callback function to ClingoControl,
which can also handle theory output, so added new options for the python/lua interface to show theory terms, whatever they are.
There is still an interface missing to add these theory terms for the python propagators, but its a first try.

What do think of it?
Do i get a merge ?
See ya on Monday,

Max


rkaminsk:
It looks like your pull request does not just contain the suggested feature but also all the csp stuff. I would like to wait with merging this. Can you prepare a branch please that just contains the new feature?

PR #5, opened 2016-06-27
Callback

MaxOstrowski:
Callback interface for ClingoControl

### DIFF
--- a/libgringo/src/python.cc
+++ b/libgringo/src/python.cc
@@ -1362,10 +1362,11 @@ places like - e.g., the main function.)";
     static PyObject *atoms(Model *self, PyObject *pyargs, PyObject *pykwds) {
         PY_TRY
             unsigned atomset = 0;
-            static char const *kwlist[] = {"atoms", "terms", "shown", "csp", "comp", nullptr};
-            PyObject *pyAtoms = Py_False, *pyTerms = Py_False, *pyShown = Py_False, *pyCSP = Py_False, *pyComp = Py_False;
-            if (!PyArg_ParseTupleAndKeywords(pyargs, pykwds, "|OOOOO", const_cast<char**>(kwlist), &pyAtoms, &pyTerms, &pyShown, &pyCSP, &pyComp)) { return nullptr; }
+            static char const *kwlist[] = {"atoms", "theory", "terms", "shown", "csp", "comp", nullptr};
+            PyObject *pyAtoms = Py_False, *pyTheory = Py_False, *pyTerms = Py_False, *pyShown = Py_False, *pyCSP = Py_False, *pyComp = Py_False;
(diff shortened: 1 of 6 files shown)


rkaminsk:
Added some comments to the commit. Just small things can you adress these before I pull please. -R

libgringo/src/python.cc
```diff
@@ -1362,10 +1362,11 @@ places like - e.g., the main function.)";
     static PyObject *atoms(Model *self, PyObject *pyargs, PyObject *pykwds) {
         PY_TRY
             unsigned atomset = 0;
-            static char const *kwlist[] = {"atoms", "terms", "shown", "csp", "comp", nullptr};
-            PyObject *pyAtoms = Py_False, *pyTerms = Py_False, *pyShown = Py_False, *pyCSP = Py_False, *pyComp = Py_False;
-            if (!PyArg_ParseTupleAndKeywords(pyargs, pykwds, "|OOOOO", const_cast<char**>(kwlist), &pyAtoms, &pyTerms, &pyShown, &pyCSP, &pyComp)) { return nullptr; }
+            static char const *kwlist[] = {"atoms", "theory", "terms", "shown", "csp", "comp", nullptr};
```
rkaminsk: Please move theory after csp and update the docstring.

rkaminsk:
Calling addToModel before creating the ClingoModel is unnecessary. The function can be called in ClingoModel::atoms as well. Also, you might want to change its signature to void addToModel(Clasp::Model const&, Gringo::SymVec&) to avoid allocating an unnessary vector.

MaxOstrowski:
On 06/28/2016 03:13 PM, rkaminsk wrote:
> Calling addToModel before creating the ClingoModel is unnecessary. The
> function can be called in ClingoModel::atoms as well.
Wont it be called every time you call atom then ?
Does not sound so good.


> Also, you might
> want to change its signature to void addToModel(Clasp::Model const&,
> Gringo::SymVec&) to avoid allocating an unnessary vector.
Could be done. I'll wait until you decided on the topic above.

Get well.
Max

MaxOstrowski:
Forget about the previous mail.
Did the changes.
Have fun.

Max

rkaminsk:
Did a couple more changes. Going to merge if the tests pass...

rkaminsk:
There is one more thing that comes to my mind: addToModel should take a "bool complement" to indicate that the complement was requested. You can still add nothing in this case. This makes the interface more general.

MaxOstrowski:
On 06/29/2016 11:25 AM, rkaminsk wrote:
> There is one more thing that comes to my mind: addToModel should take a
> "bool complement" to indicate that the complement was requested. You can
> still add nothing in this case. This makes the interface more general.

Done

rkaminsk:
Merged. You might want to remove the branch now.

PR #7, opened 2016-08-04
Change weather -> whether in libgringo/clingo.h.

Horrowind:


rkaminsk:
Yeah, I always mix these two up. And thought I already fixed this but the uppercase ones escaped me. Going to merge.

PR #11, opened 2016-11-02
Add cast for LiteralId data_.offset to fix Windows bitfield issue

thomas-ae-smith:
When clingo.dll is compiled in VS2015, the 'offset' member of the LiteralId bitfield does not get set correctly, presumably due to platform differences. This causes all LiteralIds to be created with an offset of 0; a knock-on effect of this is that if multiple PredicateLiterals are constructed only the first one is assigned a UID, as subsequent checks (lines 948,949 in libgringo/src/output/literals.cc) mistakenly see that a UID already exists (but for the first PredicateLiteral rather than the one being constructed).

This can be demonstrated by changing line 33 of app/example/main.cc to use "a(1). a(2)" as input to the grounder, and then running the example. Expected output: "a(1) a(2)". Actual output "a(1)". After the fix, output is "a(1) a(2)" as expected. 

This fix casts the 'offset' input to the LiteralId constructor to be the same width as the bitfield (uint64_t) before assigning it to the bitfield member, which appears sufficient to fix the assignment issue. (Similar issues have not been observed with the domain property and so it has been left unaltered for the sake of minimality.)

This change has not been tested with compilers other than VS2015 as I don't have access to suitable build environments.

### DIFF
--- a/libgringo/gringo/output/literal.hh
+++ b/libgringo/gringo/output/literal.hh
@@ -65,7 +65,7 @@ public:
     LiteralId()
     : repr_(std::numeric_limits<uint64_t>::max()) { }
     LiteralId(NAF sign, AtomType type, Potassco::Id_t offset, Potassco::Id_t domain)
-    : data_{static_cast<uint32_t>(sign), static_cast<uint32_t>(type), domain, offset} { }
+    : data_{static_cast<uint32_t>(sign), static_cast<uint32_t>(type), domain, static_cast<uint64_t>(offset)} { }
     Potassco::Id_t offset() const { return data_.offset; }
     Potassco::Id_t domain() const { return data_.domain; }
     AtomType type() const { return static_cast<AtomType>(data_.type); }


rkaminsk:
Hi, I cannot reproduce this with gcc/clang. On first sight, I would say this is a compiler bug. I am going to investigate and then come back to you. Thanks for reporting the problem. Looks like something that was not easy to find... -R

rkaminsk:
Actually, I also cannot reproduce the problem with VS 2015. I tested the {Release,Debug}/x86 and {Release/x64} configurations. In all cases I get the correct models. What configuration are you using? Is your VS up to date?

thomas-ae-smith:
I'm using Visual Studio Professional 2015 Version 14.0.25123.00 Update 2 (so not fully up to date); have tested with a fresh* checkout of the vs15 branch in {Release,Debug}/{x86,x64} and in all cases the model contains only a(1) and not a(2).

image

This being a compiler bug seems likely, and for the moment this cast corrects the behaviour, at least for me in Update 2. I will see if it's possible for me to upgrade to Update 3 tomorrow morning and investigate.

(* I had to add ../../libgringo to the Additional Include Directories property of the cclingo and example projects, and (to save time) I just copied across the libgringo/generated files from my existing checkout)

rkaminsk:
This is definitely a compiler bug that is fixed with Update 3. I am not sure if it is better to update Visual Studio than to apply the patch.

PS: We are currently porting clingo to use cmake. It can already generate VS projects. The code is in the cmake branch.

rkaminsk:
Hi, we brought the cmake branch in shape to be be compiled with VS15. Interestingly, there was a problem with the LiteralId class in Update 3 too. This workaround might help with your VS version, too. Note that I am going to remove the vs15 branch soon and merge the cmake branch back into the wip branch.

PR #20, opened 2016-12-24
add virtual destructor to GroundProgramObserver class

iensen:
Sometimes we need to delete instances of an observer through pointers to the base class.
Missing virtual destructor leads to an undefined behavior in such cases.

There is actually a virtual destructor in the theory propagator class, so I assume this is just an omission.


rkaminsk:
Thanks, leaving the virtual destructor out was not intended. -R

rkaminsk:
I do not want to apply the patch to the master but the wip branch instead, which is the base for the next release. The directory layout changed in this branch, which does not allow the patch to be applied without manual intervening. Since the patch is so small, I simply apply it add it by hand and close this request.

iensen:
Thanks!

PR #27, opened 2017-03-29
Add dereference operators to Optional

pluehne:
This implements four dereference operators for `Clingo::Optional`, in analogy to the spefication of [`std::optional`](http://en.cppreference.com/w/cpp/utility/optional/operator*).


rkaminsk:
Do you also need an operator->() while we are at it?

pluehne:
Which kind of `operator->` do you have in mind? [`Clingo::Optional`](https://github.com/potassco/clingo/blob/wip/libclingo/clingo.hh#L190) already has `T *operator->()` and `T const *operator->() const` if I see that correctly.

PR #29, opened 2017-03-31
Fix: Variants cannot be move-assigned from descendant Variants

pluehne:
This fixes an issue where move-assigning `Clingo::Variant`s from their own descendant `Variant`s would cause a segmentation fault.


pluehne:
Thanks! I had forgotten about covering this in a unit test case.

PR #32, opened 2017-04-11
Better support for Variants with move-only types

pluehne:
This makes it more convenient to use `Clingo::Variant` with move-only types.

Without this patch, accidentally copying variants of uncopyable types would not be noticed by the compiler already but instead result in an error at runtime.

### DIFF
--- a/libclingo/clingo.hh
+++ b/libclingo/clingo.hh
@@ -243,7 +243,7 @@ class Variant {
     using Holder = Detail::VariantHolder<1, T...>;
 public:
     Variant(Variant const &other) : Variant(other.data_) { }
-    Variant(Variant &&other)      { data_.swap(other.data_); }
+    Variant(Variant &&other) noexcept { data_.swap(other.data_); }
     template <class U>
     Variant(U &&u, typename std::enable_if<Detail::TypeInList<U, T...>::value>::type * = nullptr) { emplace2<U>(std::forward<U>(u)); }
     template <class U>
@@ -258,7 +258,7 @@ public:
     }
     ~Variant() { data_.destroy(); }
     Variant &operator=(Variant const &other) { return *this = other.data_; }
-    Variant &operator=(Variant &&other) { return *this = std::move(other.data_); }
+    Variant &operator=(Variant &&other) noexcept { return *this = std::move(other.data_); }
     template <class U>
     typename std::enable_if<Detail::TypeInList<U, T...>::value, Variant>::type &operator=(U &&u) {
         emplace2<U>(std::forward<U>(u));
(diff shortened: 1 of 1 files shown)


rkaminsk:
It looks like only gcc somehow accepts the change (which now looking at this I find surprising). Not sure if there is a solution that also works for the other compilers.

pluehne:
I see. I removed the commit in question so that you can just pull the remaining change (making all move operations `noexcept`).

Find below a patch of the commit that didn’t make it in yet for later reference.

```diff
From 53492d9878c77e8d4ae0db1ce7ef07aba5d9d911 Mon Sep 17 00:00:00 2001
From: =?UTF-8?q?Patrick=20L=C3=BChne?= <patrick-github@luehne.de>
Date: Tue, 11 Apr 2017 16:58:33 +0200
Subject: [PATCH] Properly providing copy constructor for Variant if types are
 copyable.

---
 libclingo/clingo.hh | 22 +++++++++++++++++++++-
 1 file changed, 21 insertions(+), 1 deletion(-)

diff --git a/libclingo/clingo.hh b/libclingo/clingo.hh
index 5aa3c9c..eea46ba 100644
--- a/libclingo/clingo.hh
+++ b/libclingo/clingo.hh
@@ -70,6 +70,26 @@ inline std::ostream &operator<<(std::ostream &out, TruthValue tv) {
 
 namespace Detail {
 
+template <bool... T>
+struct And {
+    constexpr static bool value = true;
+};
+
+template <bool T, bool... U>
+struct And<T, U...> {
+    constexpr static bool value = T && And<U...>::value;
+};
+
+template <bool B>
+struct DeleteCopyConstructor {
+    DeleteCopyConstructor() {}
+    DeleteCopyConstructor(const DeleteCopyConstructor &) = delete;
+};
+
+template<>
+struct DeleteCopyConstructor<true> {
+};
+
 template <class T, class... U>
 struct TypeInList : std::false_type { };
 
@@ -239,7 +259,7 @@ private:
 };
 
 template <class... T>
-class Variant {
+class Variant : Detail::DeleteCopyConstructor<Detail::And<std::is_copy_constructible<T>::value...>::value> {
     using Holder = Detail::VariantHolder<1, T...>;
 public:
     Variant(Variant const &other) : Variant(other.data_) { }
-- 
2.12.2
```

rkaminsk:
I'll merge once the tests are green and have a look for another solution to enable/disable the copy constructor later.

pluehne:
Thank you so much! No need to rush things, my anthem release is now stable, and we can look for better solutions at a later point in time 😃.

rkaminsk:
I just had to look how they did this for `std::vector`. There `std::is_copy_constructible` is always true, but copy construction will fail if the `value_type` does not support it. Doing the same for the variant is probably a good idea. Then the clingo variant is more like a container and not like a value type, which makes kind of sense the way it is implemented.

```
#include <vector>
#include <type_traits>

struct X { 
    X();
    X(X const &) = delete;
};

int main() { 
    static_assert(std::is_copy_constructible<std::vector<X>>::value, "");
    std::vector<X> x, y;
    // fails
    // x = y;
}
```

PR #85, opened 2018-02-07
Dynlib

MaxOstrowski:
Adding the possibility to retrieve the C pointer to the control object.
Can be used for dynamically loading propagators.


MaxOstrowski:
Any updates on this?

rkaminsk:
I will address this for the next release.

rkaminsk:
Added this as a _to_c property to the Control object. Since I am not a fan of adding these things to the python interface, I prefixed it with `_`. It won't appear in the documentation but this can be requested if qualified explicitly: `help(clingo.Control._to_c)`.

PR #93, opened 2018-04-08
Expose exponentiation operator via the API

pluehne:
I noticed that the Clingo API is missing the exponentiation (or ”power”) operator. In my project, I have a `switch` statement that covers all types of binary operators and noticed that it made a difference if I had a `default` case or if I listed all possible cases separately if my input program contained `**` as an arithmetic operator.

This was a bit tricky to debug, because the compiler does not give any hints about the issue, as it doesn’t expect out-of-bounds values with `enum class` values (which possibly is a case of undefined behavior).

This pull request adds the missing exponentiation operator to the C API as well as the C++ and Python bindings.

If I see that correctly, the binary operator type is transferred from Clingo’s internal representation to the C API in: https://github.com/potassco/clingo/blob/9bbad4cf18723be04d14c4d8a4aba424a29de4ea/libclingo/src/ast.cc#L137 from a Clingo-internal `BinOp`: https://github.com/potassco/clingo/blob/9bbad4cf18723be04d14c4d8a4aba424a29de4ea/libgringo/gringo/term.hh#L48 through a `static_cast`. (This also explains why the compiler doesn’t complain about the missing `enum class` value.) For this reason, it seems like it’s enough to add the missing exponentiation operator to the respective enums in the C API as well as the C++ and Python bindings. I’m not sure about the Lua bindings, but to me it looks like they don’t need to be touched in this case.

I’m not sure whether this can be unit-tested in a meaningful way. Please also let me know if I have overseen another place in the code that would need to be adjusted :slightly_smiling_face:.

### DIFF
--- a/libclingo/clingo.h
+++ b/libclingo/clingo.h
@@ -2184,7 +2184,8 @@ enum clingo_ast_binary_operator {
     clingo_ast_binary_operator_minus          = 4,
     clingo_ast_binary_operator_multiplication = 5,
     clingo_ast_binary_operator_division       = 6,
-    clingo_ast_binary_operator_modulo         = 7
+    clingo_ast_binary_operator_modulo         = 7,
+    clingo_ast_binary_operator_power          = 8
 
 };
 typedef int clingo_ast_binary_operator_t;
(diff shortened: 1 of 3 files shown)


pluehne:
@rkaminsk: Thanks for caring about the missing unit test. I wish I’d had have had the time to write one by myself :slightly_smiling_face:.

PR #108, opened 2018-05-28
Statistics

MaxOstrowski:
Modified statistics object for reading and writing statistics.


rkaminsk:
I am not finished reviewing this. One thing is that I do not like is to const_cast the statistics in the C++ interface. Here is an alternative way we could implement it:
```c++
#include <memory>
#include <iostream>

struct Impl {
    explicit Impl(char const *y) : x{y} { }
    void mutable_(char const *y) {
        x = y;
    }
    char const *immutable() const {
        return x;
    }
    char const *x;
};

class Statistics {
public:
    explicit Statistics(Impl const *impl) : impl_{impl} { }
    char const *get() const {
        return impl_->immutable();
    }
    Impl const *to_c() const {
        return impl_;
    }
private:
    Impl const *impl_;
};

class UserStatistics : public Statistics {
public:
    explicit UserStatistics(Impl *impl) : Statistics{impl} { }
    void set(char const *x) {
        to_c()->mutable_(x);
    }
    Impl *to_c() const {
        // this const_cast is safe because the UserStatistics object was
        // constructed from a non-const pointer
        return const_cast<Impl*>(Statistics::to_c());
    }
};

int main() {
    std::unique_ptr<Impl> impl = std::make_unique<Impl>("foo");
    Statistics a{impl.get()};
    UserStatistics b{impl.get()};

    std::cerr << a.get() << std::endl;
    b.set("bar");
    std::cerr << a.get() << std::endl;
    std::cerr << static_cast<Statistics&>(b).get() << std::endl;
    // this cast is valid but a good linter will give a warning
    // and also we made clear that this is not a good idea by
    // providing the interfaces as we do
    static_cast<UserStatistics&>(a).set("foo");
}
```

rkaminsk:
And in my code I put the const after the type.

MaxOstrowski:
OK with me (as always).
You wanna do it or I?

libclingo/clingo.h
```diff
@@ -1989,22 +1989,26 @@ enum clingo_statistics_type {
 //! Corresponding type to ::clingo_statistics_type.
 typedef int clingo_statistics_type_t;
 
-//! Handle for to the solver statistics.
+//! Handle for the solver statistics.
 typedef struct clingo_statistic clingo_statistics_t;
 
+//! Callback to add user defined statistics.
+typedef bool (*clingo_set_user_statistics) (clingo_statistics_t* stats, void* data);
```
rkaminsk: All typedefs in clingo.h have suffix `_t`,  callbacks have suffix `_callback_t`. I would recommend to use `clingo_statistics_update_callback_t`. Other recommendations for names are welcome.

libclingo/clingo.h
```diff
@@ -3215,7 +3256,19 @@ CLINGO_VISIBILITY_DEFAULT bool clingo_control_is_conflicting(clingo_control_t *c
...
+//!
+//! See the @ref Statistics module for more information.
+//! Several callbacks supported.
+//!
+//! @param[in] control the target
+//! @param[in] cb a callback to the set_user_statistic function
+//! @param[in] data user data passed to the callback function
+//! @see ::clingo_set_user_statistics
+//! @return whether the call was successful
+CLINGO_VISIBILITY_DEFAULT bool clingo_control_add_user_statistics(clingo_control_t *control, clingo_set_user_statistics cb, void* data);
```
rkaminsk: I do not like the name `clingo_control_add_user_statistics`. In other places I used register instead of add. What about `clingo_control_register_user_statistics`.

libclingo/clingo.hh
```diff
@@ -2101,6 +2107,7 @@ public:
     }
     Configuration configuration();
     Statistics statistics() const;
+    void add_user_statistics(UserStatisticCallback& cb);
```
rkaminsk: add/register

libclingo/clingo.h
```diff
@@ -2057,10 +2079,21 @@ CLINGO_VISIBILITY_DEFAULT bool clingo_statistics_map_subkey_name(clingo_statisti
...
+//! Create a subkey under the given name.
+//!
+//! @pre The @link clingo_statistics_type() type@endlink of the entry must be @ref ::clingo_statistics_type_map.
+//! @param[in] statistics the target statistics
+//! @param[in] key the key
+//! @param[in] name the name of the new subkey
+//! @param[in] type the type of the new subkey
+//! @param[out] subkey the resulting subkey
+//! @return whether the call was successful
+CLINGO_VISIBILITY_DEFAULT bool clingo_statistics_map_create(clingo_statistics_t *statistics, uint64_t key, char const *name, clingo_statistics_type_t type, uint64_t *subkey);
```
rkaminsk: I do not like the name `map_create`. I think `map_add_subkey` would better fit with the existing interface.

libclingo/clingo.hh
```diff
@@ -1942,12 +1944,16 @@ public:
...
     StatisticsArrayIterator begin() const;
     StatisticsArrayIterator end() const;
+    Statistics push(StatisticsType type);
     // maps
     Statistics operator[](char const *name) const;
-    Statistics get(char const *name) { return operator[](name); }
+    Statistics get(char const *name) const { return operator[](name); }
+    bool has_subkey(char const *name) const;
     StatisticsKeyRange keys() const;
+    Statistics create(char const *name, StatisticsType type);
```
rkaminsk: create/add_subkey

libclingo/clingo.hh
```diff
@@ -1942,12 +1944,16 @@ public:
...
     // leafs
     double value() const;
+    void set(double d);
```
rkaminsk: We could also provide an assignment operator for this.

rkaminsk:
This pull request is too messy. I'll continue working on it in the statistics branch in clingo.

PR #111, opened 2018-06-05
Statistics

rkaminsk:


rkaminsk:
@BenKaufmann this pull request adds support for user statistics to the clingo APIs. What I do not like about it (on the clingo-side) is that currently there is a function in the Control class to register a callback to update statistics. At the moment we already have an event handler that can be passed to the solve() function. To not end up with too many functions to register callbacks, I would prefer to add an `update_statistics` event there. Is it guaranteed from the clasp-side that statistics callbacks are fired only during solving?

BenKaufmann:
@rkaminsk 
Stats callbacks registered added via `ClaspFacade::addStatisticsCallback()` are called exactly once at the end of each solving step. So yes, they are only fired during solving.

rkaminsk:
Thanks, it should be easy then to put it in clingo's event handler.

PR #114, opened 2018-06-20
added hidden _to_c member to most handles

MaxOstrowski:
I added some _to_c members to the Python interface for all handles that can easily be converted.
I excluded types and singles (atoms, iterators, etc...).


rkaminsk:
Merged and also added a _to_c to symbols to be able to pass around those.

PR #125, opened 2018-11-03
Control.solve(): rename async keyword argument to async_

tkren:
`async` is a reserved keyword starting with python 3.7:
https://docs.python.org/3/whatsnew/3.7.html#summary-release-highlights

This commit introduces a backwards incompatible change in the clingo
python module. The `Control` class method `solve()` now uses the following
keyword argument list:

`solve(self, assumptions, on_model, on_finish, yield_, async_)`

### DIFF
--- a/libpyclingo/pyclingo.cc
+++ b/libpyclingo/pyclingo.cc
@@ -6320,7 +6320,7 @@ active; you must not call any member function during search.)";
         CHECK_BLOCKED("solve");
         Py_XDECREF(stats);
         stats = nullptr;
-        static char const *kwlist[] = {"assumptions", "on_model", "on_statistics", "on_finish", "yield_", "async", nullptr};
+        static char const *kwlist[] = {"assumptions", "on_model", "on_statistics", "on_finish", "yield_", "async_", nullptr};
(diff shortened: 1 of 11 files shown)


libpyclingo/pyclingo.cc
```diff
@@ -6320,7 +6320,7 @@ active; you must not call any member function during search.)";
         CHECK_BLOCKED("solve");
         Py_XDECREF(stats);
         stats = nullptr;
-        static char const *kwlist[] = {"assumptions", "on_model", "on_statistics", "on_finish", "yield_", "async", nullptr};
+        static char const *kwlist[] = {"assumptions", "on_model", "on_statistics", "on_finish", "yield_", "async_", nullptr};
```
rkaminsk:
```suggestion
        static char const *kwlist[] = {"assumptions", "on_model", "on_statistics", "on_finish", "yield_", "async", "async_", nullptr};
```
One could also accept both `async` and `async_` for backward compatibility. But then I also broke the API before...

rkaminsk: I cannot promise a new release by January 2019 but will keep this in mind. Right now there are only the things in the issue tracker I still want to add because some of our systems need them.

rkaminsk:
I implemented the backward compatible option based on your pull request.

PR #126, opened 2018-11-03
controller-async: Read-Eval-Print loop is broken in python3

tkren:
The `input` variable is unassigned whenever accessing `raw_input()` raises a
`NameError`.

This commit uses python's `code` module to implement input parsing.

### DIFF
--- a/examples/clingo/controller-async/controller.py
+++ b/examples/clingo/controller-async/controller.py
@@ -2,6 +2,7 @@
 
 import os
 import readline
+import code
 import atexit
 import signal
 import clingo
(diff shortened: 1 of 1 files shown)


rkaminsk:
Are you sure this is necessary? The `input` function should be available in Python 3. From the docs:
- <https://docs.python.org/3/library/functions.html#input>
- <https://docs.python.org/2/library/functions.html#raw_input>

tkren:
I forgot to mention that I've set the shebang in `controller.py` to `#!/usr/bin/env python3` (`#!/usr/bin/env python` gives `python2` on Debian).

 With this change I get:

     % ./controller.py                                    

     this prompt accepts the following commands:
       solve              - start solving
       exit/EOF           - terminate the solver
       Ctrl-C             - interrupt current search
       less_pigeon_please - select an easy problem
       more_pigeon_please - select a difficult problem

     Traceback (most recent call last):
       File "./controller.py", line 123, in <module>
         ct.run()
       File "./controller.py", line 55, in run
         line = input('> ')
     UnboundLocalError: local variable 'input' referenced before assignment

rkaminsk:
Now I see. It is because of Python's scoping rules. Adding a `global input` before the `try` also fixes the problem. I am going to add a fix for this. There are also other places where this code is used.

PS: You might want to consider to add packages for the clingo python module for different python versions. The python module can be build against against an installed clingo. I added the `CLINGO_USE_LIB` cmake option for this.

rkaminsk:
I decided to fix it a little differently. There is no need to pull in the code module.

PR #127, opened 2018-11-12
libclingo: fix catching polymorphic exception by value

timn:
GCC 8 warns about polymorphic structs being caught by value rather than by reference. Fix by catching a reference. This currently prevents building on, e.g., Fedora 28 or later.


rkaminsk:
Thanks. Maybe also catch it by const reference (because this is the standard way).

rkaminsk:
Thanks.

Issue #86, opened 2018-02-14
const arguments in the C API


sthiele:
Hi Roland,

I wonder if some function arguments in the C API could be const. For example in
https://github.com/potassco/clingo/blob/1053dee9c920bb0c3029a7f8c78a4b126afba435/libclingo/clingo.h#L549
Getting the size should not change the atoms. 
`bool clingo_symbolic_atoms_size(clingo_symbolic_atoms_t const *atoms, size_t *size)`

Best,
Sven

rkaminsk:
In the whole API I never used const for opaque types. Since symbolic_atoms is an opaque type and has a read-only interface, adding const would add no additional safety here. There are other places where this could make sense. In the end I hope that not too many people are going to use the C-interface directly. And instead use a higher-level language binding.

sthiele:
It would be inform user of the API or a program generating bindings to a higher-level language that this function does not mutate the object.

rkaminsk:
I don't think that there is much gain adding this to a C API. As I said, I would like to leave this to the people implementing higher level API's.

sthiele:
Do you accept a PR? I am working on a idiomatic Rust API. It would safe me work if this would be fixed in the C API because then I don't have to edit the auto generated bindings: https://github.com/sthiele/clingo-rs/commit/bebef2143ebc4541e82db3bb9419b1e0e1cb8939  Also it does no harm.

rkaminsk:
I take PRs. It is just that I don't see the big gain here. The C API as it is, is quite painful to use and so would be a direct binding. To get a nice rust API, you still would have to write some wrappers. The wrappers can then just hold a C pointer and present const/non-const methods.

sthiele:
I did some of the mentioned [changes](https://github.com/sthiele/clingo/commit/835c0fc8289dd1483afa0ae06e27cbb452d67841) in my fork of the 5.3.0 release .
Maybe you have a look at it before I prepare a PR.
For which branch should I make a PR?

rkaminsk:
Please do pull requests against the `wip` branch. Guten Rutsch!

PR #135, opened 2019-01-01
adding const modifiers

sthiele:
as discussed in issue [#86](https://github.com/potassco/clingo/issues/86)


rkaminsk:
I pulled your changes into [stiehle/const-mod](/potassco/clingo/tree/stiehle/const-mod). I'll do some more changes there before merging.

PR #137, opened 2019-02-22
Fixed number mismatches

bartbog:
In some documentation (and tests) the atoms that are claimed to test for
and the atoms that are actually tested for differ. Fixed those.


rkaminsk:
Thanks.

PR #163, opened 2019-07-31
prototypical implementation of add_clause in propagate init

rkaminsk:
Hi @BenKaufmann, I added a function to add clauses while initializing the propagator. Clauses are added while propagators are initialized, i.e., right after calling `clasp_->program()->endProgram()` but before `clasp_->prepare()`. As far as I can see this should be a good point to add constraints. Can you have a quick look whether this is correct?

### DIFF
--- a/app/clingo/tests/lua/add-clause-lua.lp
+++ b/app/clingo/tests/lua/add-clause-lua.lp
@@ -1,0 +1,49 @@
+#script (lua)
+
+require("clingo")
+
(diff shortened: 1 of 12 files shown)


BenKaufmann:
@rkaminsk Looks good to me.

rkaminsk:
Thanks, then I just merge it. I am also planning to do a new release. But I will still wait a couple of weeks to see if everything is working.

PR #167, opened 2019-09-06
Use 0 instead of nullptr for python's tp_print for Python 3.8

morxa:
In Python 3.8, the reserved `tp_print` slot was changed from a function
pointer to a number.  In C, there is no `nullptr`; either a 0 or NULL
casts automatically to both pointers and numbers.

Replace `nullptr` with `0` for `tp_print` to be compatible with Python
3.8.

See the [Fedora BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1743883) for more information.


rkaminsk:
Thanks.

PR #188, opened 2020-01-14
WIP: Expose type annotations in Python API

0x326:
**Work in progress. Help would be appreciated**

## Goal

Expose type annotations via a Python "stub" so that tools like mypy and PyCharm can infer whether appropriate types are passed to clingo's methods. 

## References:

- https://www.python.org/dev/peps/pep-0561/
- https://mypy.readthedocs.io/en/latest/installed_packages.html


rkaminsk:
I just learned that this exists. What I noted before was that the code completion engine I am using already uses type hints in the docstrings for semantic completion. I am using youcompleteme, which uses jedi under the hood. Ideally, it would be possible to generate the `clingo.pyi` file from the docstrings. Otherwise, maintaining this will be cumbersome.

rkaminsk:
Have you stopped working on this? I am not opposed to merging something but it needs to be maintainable. I do not want to maintain signatures in two places. Here is an idea how to generate the clingo.pyi file. Writing the clingo_abc.py file by hand seems okay to me. Otherwise, this also needs to be part of the install target.

```python
import clingo
import types
import sys


def get_sig(entry):
    return entry.__doc__.strip().splitlines()[0]


functions = []
classes = []
modules = []
other = []

for name, entry in sorted(clingo.__dict__.items(), key=lambda x: x[0]):
    if name.startswith("_") and name != "__version__":
        continue
    if isinstance(entry, type):
        classes.append((name, entry))
    elif isinstance(entry, types.BuiltinFunctionType):
        functions.append((name, entry))
    elif isinstance(entry, types.ModuleType):
        modules.append((name, entry))
    else:
        other.append((name, entry))

sys.stdout.write("from typing import *\n")
sys.stdout.write("from clingo_abc import Application, Observer, Propagator\n\n\n")

for name, entry in other:
    if isinstance(entry, clingo.Symbol):
        sys.stdout.write("{}: Symbol\n".format(name))
    elif isinstance(entry, str):
        sys.stdout.write("{}: str\n".format(name))
    else:
        sys.stderr.write("unhandled entry: {}\n".format(name))

sys.stdout.write("\n\n")

for name, entry in functions:
    sys.stdout.write("def {}: ...\n".format(get_sig(entry)))

sys.stdout.write("\n\n")

for name, entry in classes:
    pass  # TODO
```

0x326:
I am totally in favor of making this more maintainable and I'm still interested in updating it so that it can be ready for merging. However, I don't think I'll be able to work on it until the summer, since the academic year keeps me pretty busy

rkaminsk:
Let's see. I might have a look earlier. In principle, I want to make a release in the near future. To fully generate the stub file, I should also add signatures to the docstrings of some classes. For example, for the iterators and sequences.

rkaminsk:
I'll continue developing this in the `feature/mypy` branch. There is also a refined generator for the stub file in `scratch/mypy` now. If everything goes well, this will be part of the next release.

rkaminsk:
Closing this and continuing in the mentioned PR. If you like, you can give this a try. It should already work. There is just one todo left.

PR #252, opened 2020-12-04
Consider creating 5.4.1 release for Python 3.8 support

0x326:
The code base at the `v5.4.0` tag is not compatible with Python 3.8. Ubuntu 20.04 LTS users are on Python 3.8 by default (without a way to downgrade via `apt`), which means they cannot build clingo from source.

However, shortly after v5.4.0 was released, a small patch (13a896b6a762e48c2396e3dc9f2e794020f4e6e8) was made that enabled building for Python 3.8.  

Would it be possible to release the code base at that point as `v5.4.1` so that there is an "official" release downstream users can reference if they need the latest Python support?  To make it easier, I've already replicated the release procedure used in v5.4.0 (unless I've missed something).

**This PR is not intended to be merged**. Instead, amend it as necessary (with `git commit --amend`) then tag it as `v5.4.1`


rkaminsk:
I created a new release.

0x326:
@rkaminsk Thanks!

PR #259, opened 2021-01-14
Add links to changelog

0x326:


rkaminsk:
I'll see what I do about this one and keep it open for now. On the releases page for example github already links to issues without any extra work.

rkaminsk:
I prefer to keep the changes file simpler and not attach all the links.

PR #269, opened 2021-02-20
add single-shot solving option

rkaminsk:
See also potassco/clasp#64.

- [x] add `--single-shot` option to enable some optimization that cannot be applied in the multi-shot case
- [x] raise an error if more than one step is started in single-shot mode
- [x] add unit tests

### DIFF
--- a/libclingo/clingo/clingocontrol.hh
+++ b/libclingo/clingo/clingocontrol.hh
@@ -97,6 +97,7 @@ struct ClingoOptions {
     bool                          wNoOther              = false;
     bool                          rewriteMinimize       = false;
     bool                          keepFacts             = false;
+    bool                          singleShot            = false;
     Foobar                        foobar;
 };
 
@@ -386,7 +387,6 @@ public:
     bool                                                       verbose_               = false;
     bool                                                       parsed                 = false;
     bool                                                       grounded               = false;
-    bool                                                       incremental_           = true;
     bool                                                       configUpdate_          = false;
     bool                                                       initialized_           = false;
     bool                                                       incmode_               = false;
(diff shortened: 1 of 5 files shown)


ssardina:
@rkaminsk question on the `--single-shot` and the Python API: can the option be set via the configuration of the Control? I couldn't find it but maybe I am missing it? 

(sorry I follow-up here just in case somebody else find it)

rkaminsk:
The option has to be set upon creation of the control object. It cannot be set via the dynamic configuration.

ssardina:
wow, thanks for the quick reply!

> The option has to be set upon creation of the control object. It cannot be set via the dynamic configuration.

You mean via the `arguments` option of the `Control` class like this, right?

```python
control = clingo.Control(arguments="--single-shot")
```

That is all good for my own application that creates a `Control`, but I couldn't see how to pass arguments if I use teh  [`Application`](https://potassco.org/clingo/python-api/current/clingo/application.html) framework that comes with Clingo

Presumably the control object is created automatically, so is there any way of passing arguments like `--single-shot`?

rkaminsk:
> wow, thanks for the quick reply!
> 
> > The option has to be set upon creation of the control object. It cannot be set via the dynamic configuration.
> 
> You mean via the `arguments` option of the `Control` class like this, right?
> 
> ```python
> control = clingo.Control(arguments="--single-shot")
> ```
> 
> That is all good for my own application that creates a `Control`, but I couldn't see how to pass arguments if I use teh [`Application`](https://potassco.org/clingo/python-api/current/clingo/application.html) framework that comes with Clingo
> 
> Presumably the control object is created automatically, so is there any way of passing arguments like `--single-shot`?

For the application, it can be passed on the command line as with clingo.

In principle one can also inject it into the options befor calling `clingo_main`. It's a bit hacky, but might suite your purpose:
```python
args = sys.argv[1:]
if not "--single-shot" in args:
    args.append("--single-shot")
clingo_main(App(), args)
```

ssardina:
Ohh, I see... so is `clingo_main()` function the one that creates the `Control` object (and uses `args` as arguments) and passes it to `main`?

PR #271, opened 2021-02-25
Add generic multi-shot solving approach example

ingambe:
## Description

Following KR Seminar presentation of our multi-shot solving approach, we discuss the potential interest of applying the developed multi-shot technique for the Aircraft scheduling and maintenance planning problem to other problems.

To do so, a bit of clean-up was necessary, and, as suggested during the seminar, transforming the messy function into a class.

## What has been done

Inside the example folder, a new folder called **multi_shot** containing the multi-shot approach and the encoding with an instance to test it.
A small README with further information is also included.


rkaminsk:
I am bit busy at the moment but I will come back to you. Since this is not critical for a release, there is also no need to merge this quickly (if we decide to merge). Just some points I noticed glossing over this:

- clingo does not use the sum of costs to optimize but the lexicographic ordering
- the timeout handling can probably be written nicer using the asynchronous API; the way the timer is managed looks too difficult
- it would probably be nice to base this example on clingo's application class (check <https://github.com/potassco/clingo/blob/master/examples/clingo/tmode/tmode-cffi.py>)

Also development happens in the wip branch and then gets merged into the master if the code is deemed stable enough. Please base your PR against the wip branch.

ingambe:
Thank you for the detailed explanation, I will apply the modification and do another pull request against the wip branch.

rkaminsk:
> Thank you for the detailed explanation, I will apply the modification and do another pull request against the wip branch.

Just change the target of this branch and continue working here. No need to open a new PR.

PR #281, opened 2021-03-15
Remove pyclingo

rkaminsk:


alalazo:
Skimming through the PR, it seems this is dropping the support for Python 2 and making CFFI mandatory for Python extensions - am I correct on that?

rkaminsk:
Yes, cffi and at least Python 3.6 are required now. In principle cffi supports Python 2 but I really wanted to use Python's type annotations without any crude workarounds. Also, all Python versions before 3.6 are end of life and even 3.6 will only be supported until the end of this year.

PR #293, opened 2021-04-16
add __repr__ to every item that has __str__

MaxOstrowski:
Draft PR, do not merge.

Question 1:  Do theory related __repr__ make sense, as they can't be created by hand ?
Question 2: Should a single test file for every class be added currently simply testing the repr function ?

### DIFF
--- a/libpyclingo/.pylintrc
+++ b/libpyclingo/.pylintrc
@@ -1,0 +1,582 @@
+[MASTER]
+
+# A comma-separated list of package or module names from where C extensions may
+# be loaded. Extensions are loading into the active Python interpreter and may
+# run arbitrary code.
+extension-pkg-whitelist=
+
+# Add files or directories to the blacklist. They should be base names, not
+# paths.
+ignore=CVS
+
+# Add files or directories matching the regex patterns to the blacklist. The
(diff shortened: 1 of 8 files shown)


rkaminsk:
The important classes that need a repr:

- `clingo.ast.AST`
- `clingo.symbol.Symbol`

All other classes are constructed from some internal representation and their representation is not much better than their Python representation. Only for debugging purposes it might be helpful to see how they were constructed. So we can still add them but it is not that helpful.

MaxOstrowski:
I agree that it does not make sense for theory atoms at all, I will have to remove it later.
I added an example for the AST. Is this a way you could imagine to do this ? Then I would try to add all other functions in the ast.py file.

rkaminsk:
You should be able to use `_lib.g_clingo_ast_constructors.constructor` to create the representation. Just have a look at the child_keys function. Please do not generate a giant switch.

rkaminsk:
The repr function can actually be as simple as this:

```python
    def __repr__(self):
        name = str(self.ast_type).replace('ASTType', 'ast')
        args = ', '.join(repr(x) for x in self.values())
        return f'{name}({args})'
```

PR #300, opened 2021-04-22
Build PPC wheels for releases

### LINKED ISSUE #295: PowerPC wheels through manylinux
Hi,

I noticed in [this comment](https://github.com/potassco/clingo/pull/255#issue-552175531) that ppc builds for releases were being considered but they do not appear to have been a part of the 5.5.0 release.  Would it be possible to enable these?  I would be happy to contribute a PR that modifies [the GH actions script](https://github.com/potassco/clingo/blob/master/.github/workflows/manylinux.yml) to only run the ppc manylinux image when the `wip` option is false.

Thanks!

joshessman-llnl:
Resolves #295 

Continuing the discussion on #295, this PR adds PPC to the manylinux "matrix" and sets it to a no-op for non-release builds.
This required a small patch to the build system requirements to avoid building CMake and its dependencies through QEMU, which would hit the Actions 6-hour time limit.

The wip build workflow was tested here: https://github.com/joshessman-llnl/clingo/actions/runs/774538046
It failed during the PyPI stage (as I didn't have the auth token, of course), but successfully built the x86 wheels and skipped the ppc wheels.

The release build workflow was tested here: https://github.com/joshessman-llnl/clingo/actions/runs/774560389
Again the PyPI upload failed but both sets of wheels built as expected (PPC still in progress).

I also had to default to the GNU Make generator when building re2c - Ninja gets installed later in the build process.  I believe that building re2c for releases is technically unnecessary but wanted to avoid the implicit requirement that all PPC builds are releases.

### DIFF
--- a/.github/manylinux.py
+++ b/.github/manylinux.py
@@ -96,7 +96,7 @@ def run():
         mkdir('re2c_source')
         check_call(['curl', '-LJ', '-o', 're2c.tar.gz', 'https://github.com/skvadrik/re2c/archive/2.0.3.tar.gz'])
         check_call(['tar', 'xzf', 're2c.tar.gz', '-C', 're2c_source', '--strip-components=1'])
-        check_call(['cmake', '-G', 'Ninja', '-Hre2c_source', '-Bre2c_build', '-DRE2C_BUILD_RE2GO=OFF'])
+        check_call(['cmake', '-Hre2c_source', '-Bre2c_build', '-DRE2C_BUILD_RE2GO=OFF'])
         check_call(['cmake', '--build', 're2c_build', '--target', 'install'])
     else:
         check_call(['yum', 'install', '-y', 're2c', 'bison'])
--- a/.github/ppc.patch
+++ b/.github/ppc.patch
@@ -1,0 +1,8 @@
+diff --git a/pyproject.toml b/pyproject.toml
+index 7533528..962020e 100644
+--- a/pyproject.toml
++++ b/pyproject.toml
+@@ -1,2 +1,2 @@
+ [build-system]
+-requires = ["setuptools", "wheel", "scikit-build", "cmake", "ninja"]
++requires = ["setuptools", "wheel", "scikit-build", "ninja"]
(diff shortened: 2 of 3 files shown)


rkaminsk:
I already took care of removing cmake as a dependency and addressed the ninja issue in the wip branch, where clingo development happens before things go into the master. Can you please change this PR to the wip branch?

> Continuing the discussion on #295, this PR adds PPC to the manylinux "matrix" and sets it to a no-op for non-release builds.
> This required a small patch to the build system requirements to avoid building CMake and its dependencies through QEMU, which would hit the Actions 6-hour time limit.

Not exactly a nice solution because it unnecessarily launches virtual machines but probably the least complicated one there is. Let's hope the include/exclude configuration can at some point be made conditional.

> I also had to default to the GNU Make generator when building re2c - Ninja gets installed later in the build process. I believe that building re2c for releases is technically unnecessary but wanted to avoid the implicit requirement that all PPC builds are releases.

This is a good idea because like this I can test before releasing anything.

joshessman-llnl:
> I already took care of removing cmake as a dependency and addressed the ninja issue in the wip branch, where clingo development happens before things go into the master. Can you please change this PR to the wip branch?

Should be all switched over now, but I didn't see any changes in the wip <-> master diff that would allow the use of Ninja for the re2c build - possibly I'm missing something?

> Not exactly a nice solution because it unnecessarily launches virtual machines but probably the least complicated one there is. Let's hope the include/exclude configuration can at some point be made conditional.

Yep, not exactly ideal (another consideration is that the presence of a ppc job in the UI could give the impression that PPC wheels are being built) but it's simple; hopefully GitHub doesn't restrict its CI for OSS projects such that the extra time it burns starts to become a problem.

rkaminsk:
> Should be all switched over now, but I didn't see any changes in the wip <-> master diff that would allow the use of Ninja for the re2c build - possibly I'm missing something?

I think I missed to add this. I'll just merge your change.

> Yep, not exactly ideal (another consideration is that the presence of a ppc job in the UI could give the impression that PPC wheels are being built) but it's simple; hopefully GitHub doesn't restrict its CI for OSS projects such that the extra time it burns starts to become a problem.

This is always the danger with using free services. I hope they won't stop to offer this because this is a great service for all free projects. At least the scripts should be relatively easy to migrate to other services if it ever becomes necessary.

Thanks!

PR #363, opened 2022-03-13
Fix incremental projection when backend is used

rkaminsk:


libgringo/tests/output/incremental.cc
```diff
@@ -52,30 +52,35 @@ std::string iground(std::string in, int last = 3) {
     prg.rewrite(defs, module.logger);
     prg.check(module.logger);
     //std::cerr << prg;
-    // TODO: think about passing params to toGround already...
+    auto ground = [&](std::set<Sig> const &sigs, Ground::Parameters const &params) {
+        auto gPrg = prg.toGround(sigs, out.data, module.logger);
+        gPrg.prepare(params, out, module.logger);
+        gPrg.ground(context, out, module.logger);
+
```
rkaminsk, line 59:
```suggestion
```

libgringo/src/ground/program.cc
```diff
@@ -123,6 +123,9 @@ void Program::ground(Parameters const &params, Context &context, Output::OutputB
         }
     }
     for (auto &x : doms) { x->nextGeneration(); }
+
+}
```
rkaminsk, line 127:
```suggestion
}

```

PR #381, opened 2022-06-28
add theory functionality to backend

rkaminsk:


rkaminsk:
@wanko, you should be able to use this for your htc related project.

PR #382, opened 2022-07-22
Use EXPORTED_RUNTIME_METHODS

domoritz:
EXTRA_EXPORTED_RUNTIME_METHODS is deprecated. See https://emscripten.org/docs/introducing_emscripten/release_notes.html.


rkaminsk:
Thanks, I also adjusted the makefile a bit to also take care of compiling lua.

domoritz:
Sounds good. I may need to adjust https://github.com/domoritz/clingo-wasm a bit when the release comes out since I currently use Lua 5.3.6.

PR #441, opened 2023-07-07
rework hash_combine and hash_mix for performance

trws:
This ended up being a small change, tweaking the hash_combine function to work more like the boost version and selecting hash_mix based on size of size_t rather than the argument.  The end result is that for spack inputs 5.6.2 performs almost exactly as well as 5.5.2, which is a 30%+ speedup in some tests.

### DIFF
--- a/libgringo/gringo/utility.hh
+++ b/libgringo/gringo/utility.hh
@@ -184,11 +184,10 @@ bool is_value_equal_to(T const &a, T const &b);
 
 // {{{1 declaration of helpers to calculate hashes
 
-template <class T>
-T hash_mix(T const &v);
+size_t hash_mix(size_t seed);
 
-template <class T, class U, class H=std::hash<T>>
-void hash_combine(U& seed, T const &v, H h=H{});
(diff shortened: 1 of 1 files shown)


rkaminsk:
> This ended up being a small change, tweaking the hash_combine function to work more like the boost version and selecting hash_mix based on size of size_t rather than the argument. The end result is that for spack inputs 5.6.2 performs almost exactly as well as 5.5.2, which is a 30%+ speedup in some tests.

I am puzzled. As far as I understand hash functions, this should be a really bad one. Now `hash_combine` is even commutative. I'll have to dig a little deeper to see what is actually happening here.

rkaminsk:
I just played around a bit and the key element of this patch seems to be applying the hash_mix function in hash_combine. With this we can also get the same speed up with a small twist to the original implementation. I'll commit a suggestion also cleaning up the mess with the templates I introduced a bit. :wink:

rkaminsk:
If this works for you, I would merge it like this. And thanks for digging into this. I don't think I would have found it myself.

PR #449, opened 2023-08-12
Hide more symbols in ELF shared libs

haampie:


rkaminsk:
See @447 for a discussion.

PR #464, opened 2023-11-02
Replace distutils with setuptools/sysconfig

adamjstewart:
Distutils was deprecated in Python 3.10 and removed in Python 3.12.[^1] This PR removes all uses of distutils to fix Python 3.12 support.[^2]

@haampie 

[^1]: https://peps.python.org/pep-0632/
[^2]: https://setuptools.pypa.io/en/latest/deprecated/distutils-legacy.html

### DIFF
--- a/cmake/python-site.py
+++ b/cmake/python-site.py
@@ -7,10 +7,7 @@ if sys.version_info >= (3, 11):
 else:
     NEW_STYLE = False
     from site import USER_SITE
-    try:
-        from setuptools.sysconfig import get_python_lib, get_config_vars
-    except ImportError:
-        from distutils.sysconfig import get_python_lib, get_config_vars
+    from distutils.sysconfig import get_python_lib, get_config_vars
 
 parser = ArgumentParser()
 if sys.version_info >= (3, 7):
--- a/scratch/setup.py
+++ b/scratch/setup.py
@@ -1,4 +1,4 @@
-from distutils.core import setup, Extension
+from setuptools import setup, Extension
 import os
 
 path = os.path.normpath(os.path.abspath(os.path.join(os.getcwd(), "build", "release", "bin")))


cmake/python-site.py
```diff
@@ -1,11 +1,11 @@
-try:
-    from setuptools.sysconfig import get_python_lib, get_config_vars
```
adamjstewart, line 2: This code does not work, `setuptools.sysconfig` is not a real thing. This would always result in an ImportError and skip to distutils, even if setuptools is installed.

cmake/python-site.py
```diff
@@ -1,11 +1,11 @@
...
-except ImportError:
-    from distutils.sysconfig import get_python_lib, get_config_vars
 import sys
+from sysconfig import get_config_vars, get_paths
 
 if sys.argv[1] == "prefix":
-    print(get_python_lib(True, False, sys.argv[2] if len(sys.argv) > 2 else None))
```
adamjstewart, line 8: There is no exact equivalent of `get_python_lib` in sysconfig. The solution chosen here should work most of the time, but may encounter issues with weird system Python installations. For example, this doesn't work using system Python on macOS. Pip faces the same problem: https://github.com/pypa/pip/blob/22.1/src/pip/_internal/locations/__init__.py#L486

rkaminsk:
Thanks for the PR Adam. Actually, development happens in the wip branch of clingo where this issue (plus a few more)  have already been fixed. It should be ready for Python 3.12.

adamjstewart:
Interesting. Well the wip branch has the same bug, so I'll submit the PR for that branch instead.

rkaminsk:
> Interesting. Well the wip branch has the same bug, so I'll submit the PR for that branch instead.

It should not. Check the `sys.version_info >= (3, 11)` part.

scratch/setup.py
```diff
@@ -1,4 +1,4 @@
-from distutils.core import setup, Extension
```
rkaminsk, line 1: Are you using this? The scratch space is kind of a dump for ideas that I never care to clean up. :smile:

adamjstewart:
It works on 3.12, but the else block still contains the same bug and is using distutils instead. Is there any reason not to always use sysconfig?

scratch/setup.py
rkaminsk: I only use the toplevel setup.py for deploying packages.

adamjstewart: Nope, just clearing out imports. Probably worth deleting and hiding in `.gitignore`

haampie:
The idea of showing users a clean and stable `master` branch as a landing page on Github sounds like a good idea, except that it is actually not very friendly for open source contributors, who do not know this `wip` branch exists.

I would suggest (again) to make the development branch the default, cause I've had this issue too: open a PR, oops wrong branch, close, rebase, open a new PR... it's a bit of a time waste. Also I backported the distutils workaround from the main branch to older clingo, only to realize it was broken, and the actual fix was only on the hidden wip branch.

So, it's giving more problems than it solves, people know how to click "releases" on Github for something stable.

rkaminsk:
I had a look at the `setuptools.sysconfig` issue and it really dose not seem to exist. Maybe this was meant to be `setuptools._distutils.sysconfig` at some point.

rkaminsk:
> The idea of showing users a clean and stable `master` branch as a landing page on Github sounds like a good idea, except that it is actually not very friendly for open source contributors, who do not know this `wip` branch exists.
> 
> I would suggest (again) to make the development branch the default, cause I've had this issue too: open a PR, oops wrong branch, close, rebase, open a new PR... it's a bit of a time waste. Also I backported the distutils workaround from the main branch to older clingo, only to realize it was broken, and the actual fix was only on the hidden wip branch.
> 
> So, it's giving more problems than it solves, people know how to click "releases" on Github for something stable.

I still like the idea to have the master as the default. However, I also don't want to annoy contributors. For now, I'll add something to the readme and also put a PR template in place asking to use the wip branch as a base.

rkaminsk:
> It works on 3.12, but the else block still contains the same bug and is using distutils instead. Is there any reason not to always use sysconfig?

The reason is that it is working with distutils until Python 3.10. Clingo is build for quite a couple of platforms and the fine-tuning is a nightmare. Also functions like `get_preferred_scheme` are only available from python 3.10 onwards and would need a workaround.

It really looks like the `try` import block could be removed.

haampie:
> > The idea of showing users a clean and stable `master` branch as a landing page on Github sounds like a good idea, except that it is actually not very friendly for open source contributors, who do not know this `wip` branch exists.
> > I would suggest (again) to make the development branch the default, cause I've had this issue too: open a PR, oops wrong branch, close, rebase, open a new PR... it's a bit of a time waste. Also I backported the distutils workaround from the main branch to older clingo, only to realize it was broken, and the actual fix was only on the hidden wip branch.
> > So, it's giving more problems than it solves, people know how to click "releases" on Github for something stable.
> 
> I still like the idea to have the master as the default. However, I also don't want to annoy contributors. For now, I'll add something to the readme and also put a PR template in place asking to use the wip branch as a base.

Thank you!

rkaminsk:
Thanks.

PR #490, opened 2024-03-19
Simplify clasp text output customization.

BenKaufmann:
* Update clasp and simplify `CustomTextOutput`

### DIFF
--- a/libclingo/clingo/clingo_app.hh
+++ b/libclingo/clingo/clingo_app.hh
@@ -52,7 +52,7 @@ protected:
 
     ProblemType getProblemType() override;
     void        run(Clasp::ClaspFacade& clasp) override;
-    ClaspOutput* createOutput(ProblemType f) override;
+    ClaspOutput* createTextOutput(const Clasp::Cli::ClaspAppBase::TextOptions& options) override;
     void        printHelp(const Potassco::ProgramOptions::OptionContext& root) override;
     void        printVersion() override;
 
--- a/libclingo/src/clingo_app.cc
+++ b/libclingo/src/clingo_app.cc
@@ -111,80 +111,37 @@ Clasp::ProblemType ClingoApp::getProblemType() {
     return Clasp::ClaspFacade::detectProblemType(getStream());
 }
 
-// TODO: the code below is annoying. There is too much copy and paste. The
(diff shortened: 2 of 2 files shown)


rkaminsk (commented):
Thanks for taking this up. However, I think it is not quite the same. This would break systems, which want to print something after clasp printed its model. Now they can only print something before if they want to use the default printer. Systems like clingcon or clingo-lpx won't be affected because there I took over printing completely but it is still a break.

BenKaufmann:
> However, I think it is not quite the same. This would break systems, which want to print something after clasp printed its model. Now they can only print something before if they want to use the default printer. Systems like clingcon or clingo-lpx won't be affected because there I took over printing completely but it is still a break.

Hm, I'm obviously missing something then. 

From my understanding, both versions should do the same.

- Both the old and the new code only use `CustomTextOutput` if there is a printer, i.e. `app_->has_printer()` is true.
- In the **old code**, `CustomTextOutput` directly replaced the base classe's `printModel()` function (with a nearly verbatim copy) and:
  1.  only forwarded to the base version if there's no control object. Otherwise,
  2. checked whether the model should be printed and if so **first** invoked the printer and then **afterwards** in the (optionally invoked) callback forwarded to `printValues()` of the base class. 
  3. Finally and only if `x == optQ()`, `printMeta()` of the base class was invoked
- In the **new code**, the base class `TextOutput` gained a new virtual function `printModelValues()`, which is called from `TextOutput::printModel()` only **after** the "should we print this model" check returned true.
  1. CustomTextOutput` overrides this function and again **first**  invokes the printer. 
  2. Only if the printer invokes the callback, `CustomTextOutput` then calls `printModelValues()` of the base class, which (as previously) calls `printValues()`.
  3. Finally and only if `x == optQ()`, `printMeta()` is invoked from the base class implementation of `printModel()`.

Could you please help me figure out where I went wrong?

BenKaufmann:
@rkaminsk  Sorry for the noise. Postponing the call to `printModelValues()` to after the callback has finished is already a breaking change  :disappointed:

rkaminsk:
Now it looks like it was before sans the copied code. I think it can be merged like this.

rkaminsk (approved):

PR #521, opened 2024-09-30
Fix cmake command line examples

marcbutler:
Correct cmake source directory option in examples.

### Checklist
[x] Have you followed the guidelines in our [CONTRIBUTING.md](/potassco/clingo/blob/master/CONTRIBUTING.md) document?
[x] Have you checked to ensure there aren't other open Pull Requests for the same update/change?
[x] Have you made sure that all tests succeed?

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -59,7 +59,7 @@ source releases, you can skip this step):
 To build gringo, clingo, and reify in their default configurations in release
 mode, run:
 
-    cmake -H<SOURCE_DIR> -B<BUILD_DIR> -DCMAKE_BUILD_TYPE=Release
+    cmake -S<SOURCE_DIR> -B<BUILD_DIR> -DCMAKE_BUILD_TYPE=Release
     cmake --build <BUILD_DIR>
 
 The resulting binaries and shared libraries will be in `<BUILD_DIR>/bin` and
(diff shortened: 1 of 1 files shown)


BenKaufmann:
I'd suggest being careful with this change. 
While `-S` is meanwhile the correct option, it is only available since cmake version >= 3.13. On the other hand, the undocumented `-H` option is still available and was already supported in old versions like 3.1. 

So, if we switch to `-S`, we should probably also update the minimal cmake version.

marcbutler:
> I'd suggest being careful with this change. While `-S` is meanwhile the correct option, it is only available since cmake version >= 3.13. On the other hand, the undocumented `-H` option is still available and was already supported in old versions like 3.1.

Fair point. I found it confusing as `-H` is documented as help.

marcbutler:
I have no idea why appveyor failed: there was no functional change.

rkaminsk:
> I have no idea why appveyor failed: there was no functional change.

It's unrelated to this PR. @BenKaufmann do you know what that could be?

```
/cygdrive/c/projects/clingo/clasp/libpotassco/tests/test_string_convert.cpp:371: FAILED:
  REQUIRE( std::strcmp(str.c_str(), "12.34 12.34") == 0 )
with expansion:
  -2 == 0
```

rkaminsk:
Otherwise, I think that we can update the docs for building with `-S` and `-B`. We are still compatible with cmake 3.1. Users will just have to adapt the options.

BenKaufmann:
> It's unrelated to this PR. @BenKaufmann do you know what that could be?
> 
> ```
> /cygdrive/c/projects/clingo/clasp/libpotassco/tests/test_string_convert.cpp:371: FAILED:
>   REQUIRE( std::strcmp(str.c_str(), "12.34 12.34") == 0 )
> with expansion:
>   -2 == 0
> ```

I recently re-enabled the "double parsing is locale-independent" test, which also seems to fail on cygwin. 
From what I can see, it looks like on cygwin some locale name string that is valid for `setlocale` is not valid for `std::locale`. As a result, the global locale is not reset again correctly triggering further test failures later. 

I have to investigate further but in any case, the test should at least always restore the global state.

BenKaufmann:
@rkaminsk Cygwin issue is fixed in clasp/libpotassco branch.

marcbutler:
> Otherwise, I think that we can update the docs for building with `-S` and `-B`. We are still compatible with cmake 3.1. Users will just have to adapt the options.

Would you like the PR to update the doc to indicate that the cli examples are for CMake 3.13 or later?

rkaminsk:
> Would you like the PR to update the doc to indicate that the cli examples are for CMake 3.13 or later?

No, it's okay like this.

rkaminsk:
@marcbutler, thanks for caring. :smile:

marcbutler:
@rkaminsk :smile:

PR #523, opened 2024-10-23
Add support for getting last computed model.

BenKaufmann:
Extend C and Python-API for getting last computed model after solving has finished.

With this, problematic code like:
```
with ctrl.solve(yield_=True) as sh:
    model: Model | None = None
    for model in sh:
        ...
    model.symbols(atoms=True)
```
can be safely written as:
```
with ctrl.solve(yield_=True) as sh:
    for model in sh:
        ...
    sh.last().symbols(atoms=True)
```
or
```
def do_stuff_with_last_model(m):
  ...

ctrl.solve(on_last=do_stuff_with_last_model)
```

### DIFF
--- a/app/pyclingo/_clingo.c
+++ b/app/pyclingo/_clingo.c
@@ -14254,6 +14254,59 @@ _cffi_f_clingo_solve_handle_get(PyObject *self, PyObject *args)
 #  define _cffi_f_clingo_solve_handle_get _cffi_d_clingo_solve_handle_get
 #endif
 
+static _Bool _cffi_d_clingo_solve_handle_last(clingo_solve_handle_t * x0, clingo_model_t const * * x1)
+{
+  return clingo_solve_handle_last(x0, x1);
+}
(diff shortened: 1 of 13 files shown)


rkaminsk:
This looks nice and can probably be merged as is. I'll do one last check on Friday.

PS: And thanks for fixing all the typos.

daveraja:
Thanks this looks great!

rkaminsk:
@BenKaufmann, just one small question. What happens if the search is interrupted? I am asking because people might be interested in the last (potentially sub-optimal) model model when minimizing. Is this also set then?

rkaminsk:
All the changes look fine and I would merge as is.

rkaminsk (approved):

BenKaufmann:
> What happens if the search is interrupted? I am asking because people might be interested in the last (potentially sub-optimal) model model when minimizing. Is this also set then?

Yes. The last reported model (if any) is also available when search is interrupted as this also counts as "search has finished".

Issue #525, opened 2024-11-20
Projective Solution Enumeration via Python API


thojat:
Hi,

Is there a way to activate projective solution enumeration and to add/retract project statements after grounding over the clingo/python API?

For example, I would like to generate all solutions (or brave consequences) for different sets of projected atoms and would like to avoid re-grounding or generating duplicate solutions w.r.t. the projected atoms.

I think one way to do this is to add a new lp file with project statements to the control object. But unfortunately there is no way to remove the project statements without grounding everything from scratch.

I think it would be a nice feature being able to pass a list of symbols to project to to clingo over the API after grounding, similar to assumptions or externals.

rkaminsk:
@thojat, the required functionality is available in the latest wip branch and development packages.

PR #527, opened 2024-11-26
For discussion: Add API for removing minimize constraints and updating projection atoms

BenKaufmann:
This PR is inspired by [Issue 525: Projective Solution Enumeration via Python API](https://github.com/potassco/clingo/issues/525).

It extends clingo's C and Python-API with functions for removing minimize constraints and adding/replacing sets of projection atoms.


rkaminsk:
Thanks, I'll have a look next week.

PR #530, opened 2024-12-10
Improve support for writing checkpoints in aspif format

rkaminsk:
This PR adds support to write checkpoints in aspif format, which can afterward be read again using a dedicated function.

```python
from clingo.control import Control, BackendType

# create cp1
ctl = Control()
ctl.register_backend(BackendType.Aspif, "cp1.aspif", True)
ctl.add("1 {a; b} 1.")
ctl.ground()
ctl.solve()

# create cp2
ctl = Control()
# can alternatively use load_aspif
ctl.load("cp1.aspif")
# the backend is deliberately registered later here to avoid dumping cp1 again
ctl.register_backend(BackendType.Aspif, "cp2.aspif", True)
ctl.add("c :- a. d :- b.")
ctl.ground()
ctl.solve()

# load cp1 and cp2 combined
ctl = Control(["0"])
# function load cannot be used here
ctl.load_aspif(["cp1.aspif", "cp2.aspif"])
print(ctl.solve(on_model=print))
```
The output is
```
a c
b d
SAT
```

@teiesti and @thojat, let me know what you think of this...


libclingo/clingo.h
```diff
@@ -2811,6 +2826,20 @@ CLINGO_VISIBILITY_DEFAULT void clingo_control_free(clingo_control_t *control);
 //! - ::clingo_error_runtime if parsing or checking fails
 CLINGO_VISIBILITY_DEFAULT bool clingo_control_load(clingo_control_t *control, char const *file);
 
+//! Load files in aspif format.
+//!
+//! This function should be called on an empty control object.
+//!
+//! If more than one file is given, they are merged into one file. Only the first one should have a preamble.
+//!
+//! @param[in] control the target
```
BenKaufmann, line 2835: Nitpick: there is a mismatch between the parameter name here and the one of the function (`ctl`)

libclingo/src/control.cc
```diff
@@ -2196,6 +2201,29 @@ extern "C" bool clingo_control_register_observer(clingo_control_t *control, clin
     GRINGO_CLINGO_CATCH;
 }
 
+extern "C" bool clingo_control_register_backend(clingo_control_t *control, clingo_backend_type_t type, char const *file, bool replace) {
+    GRINGO_CLINGO_TRY { 
+        auto out = gringo_make_unique<std::ofstream>(file);
+        auto backend = UBackend{}; //Output::make_backend(out, OutputFormat::reify, false, false);
```
BenKaufmann, line 2207: Nitpick: consider removing the commented-out code.

libclingo/src/gringo_app.cc
```diff
@@ -163,6 +163,17 @@ struct IncrementalControl : Control, private Output::ASPIFOutBackend {
         parser.pushFile(std::string(filename), logger_);
         parse();
     }
+    void load_aspif(Potassco::Span<char const *> files) override {
+        for (auto it = end(files), ib = begin(files); it != ib; --it) {
```
BenKaufmann, line 167: Nitpick: this (mostly) duplicated code is inconsistent regarding the use of begin/end. At least, I'd suggest dropping the (unnecessary) using declarations from the previous version.

libclingo/src/control.cc
```diff
@@ -2196,6 +2201,29 @@ extern "C" bool clingo_control_register_observer(clingo_control_t *control, clin
...
+        switch (type & 0xFFFFFFFC) {
+            case clingo_backend_type_e::clingo_backend_type_reify: {
+                backend = Output::make_backend(std::move(out), Output::OutputFormat::REIFY, (type & 1) == 1, (type &2) == 2);
```
BenKaufmann, line 2210: Nitpick: inconsistent spacing

libclingo/src/control.cc
```diff
@@ -2196,6 +2201,29 @@ extern "C" bool clingo_control_register_observer(clingo_control_t *control, clin
     GRINGO_CLINGO_CATCH;
 }
 
+extern "C" bool clingo_control_register_backend(clingo_control_t *control, clingo_backend_type_t type, char const *file, bool replace) {
+    GRINGO_CLINGO_TRY { 
+        auto out = gringo_make_unique<std::ofstream>(file);
```
BenKaufmann, line 2206: Would it make sense to check `out->is_open()` at this point already?

libclingo/clingo.h
```diff
@@ -3100,6 +3129,17 @@ CLINGO_VISIBILITY_DEFAULT bool clingo_control_theory_atoms(clingo_control_t cons
...
+//! Register a backend with the control object.
+//!
+//! This function is similar to clingo_control_register_observer except that
+//! clingo's internal backends can be selected.
+//!
+//! @param[in] control the target
+//! @param[in] type the kind of backend to register
+//! @param[in] file the file to write the result to
+//! @param[in] replace just pass the grounding to the backend but not the solver
+//! @return whether the call was successful
```
BenKaufmann, line 3141: Shouldn't this mention similar error codes as `clingo_control_load_aspif` (or at least clingo_error_bad_alloc)?

libpyclingo/clingo/control.py
```diff
@@ -660,6 +699,46 @@ def register_observer(self, observer: Observer, replace: bool = False) -> None:
...
+        """
+        Registers a backend of the given type.
+
+        This function is similar to register_observer but instead registers a predefined backend.
+
+        Parameters
+        ----------
+        type
+            The type of backend to register.
+        replace
```
BenKaufmann, line 719: Consider documenting `file`.

thojat:
I think this adds a lot of convenience for reusing already grounded programs. In the future this feature might also become an easy way to retract programs added with multishot solving, e.g., 

```python
ctl.ground([("base", [])], checkpoint=“checkpoint1”)
ctl.ground([(“p1”, [])], checkpoint=“checkpoint2”)
ctl.solve()
ctl.restore(checkpoint=“checkpoint1”)
```

As far as I understand the backend is basically an observer that generates the aspif file during grounding, right? Do you think in the future it might be possible to also store the nogoods learned during solving?

libclingo/clingo.h
rkaminsk: Fixed.

libclingo/src/control.cc
rkaminsk: Fixed.

libclingo/src/gringo_app.cc
rkaminsk: It does not compile without the using declarations. I admit, I did not check why.

libclingo/src/control.cc
rkaminsk: I don't even see where the inconsistent spacing is anymore. In another project I switched to 100% relying on clang-format. I should setup clang-format at some point but will leave it out of this PR.

libclingo/src/control.cc
rkaminsk: Yes, done.

libpyclingo/clingo/control.py
rkaminsk: Done.

libpyclingo/clingo/control.py
```diff
@@ -660,6 +699,46 @@ def register_observer(self, observer: Observer, replace: bool = False) -> None:
...
+            Whether to reify sccs.
+        reify_steps
+            Whether to reify steps.
+
+        See Also
+        --------
+        clingo.backend
+        """
+        bitset = type.value
+        if reify_sccs:
```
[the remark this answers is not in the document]
rkaminsk: I don't think it should be checked in python. But I added a check to the C-API.

libclingo/src/gringo_app.cc
rkaminsk: (The duplication will stay there for clingo series 5 but it will not stay like this in the upcoming 6.)

rkaminsk:
> I think this adds a lot of convenience for reusing already grounded programs. In the future this feature might also become an easy way to retract programs added with multishot solving, e.g.,
> 
> ```python
> ctl.ground([("base", [])], checkpoint=“checkpoint1”)
> ctl.ground([(“p1”, [])], checkpoint=“checkpoint2”)
> ctl.solve()
> ctl.restore(checkpoint=“checkpoint1”)
> ```

Restoring the same control object to a previous state is difficult. Especially, if there shall be guarantees regarding determinism, propagators, etc. The idea here was to at least make writing and loading aspif easy.

> As far as I understand the backend is basically an observer that generates the aspif file during grounding, right?

Yes, it uses the same classes that clingo uses internally.

> Do you think in the future it might be possible to also store the nogoods learned during solving?

To some extend it is already possible to store/restore learned nogoods. There are clasp's lemma-in and lemma-out options. However, this has to be used with care. The options can only be used if the underlying pogram is **exactly** the same.

rkaminsk:
If you want perfect checkpoint/restore functionality, you might want to consider <https://criu.org>. It should run on any linux system.

rkaminsk:
Going to merge. In case that further refinements are necessary let's just open a new issue.

PR #549, opened 2025-05-28
fix compilation under window

rkaminsk:
@BenKaufmann just for your info: here is the work in progress to get things running on windows

### DIFF
--- a/lib/c-api/src/ast_yaml.cc
+++ b/lib/c-api/src/ast_yaml.cc
@@ -1,5 +1,10 @@
 #include <clingo/ast.h>
 
+#ifdef _WIN32
+extern "C" auto clingo_ast_type_info_yaml() -> char const * {
+    return "not available";
+}
+#else
 extern "C" auto clingo_ast_type_info_yaml() -> char const * {
     return R"yaml(unary_operator:
   type: enum
@@ -1194,3 +1199,4 @@ statement_comment:
       doc: The type of the comment.
 )yaml";
 }
+#endif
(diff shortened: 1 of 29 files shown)


lib/c-api/src/ast_yaml.cc
```diff
@@ -1,5 +1,10 @@
 #include <clingo/ast.h>
 
+#ifdef _WIN32
```
BenKaufmann, line 3: You can get the yaml by splitting the string literal into three.
I.e. something like:
```cpp
R"yaml(unary_operator:
...
)yaml"
R"yaml(term_unary_operation:
...
)yaml"
...
```

lib/python-api/src/base.cc
```diff
@@ -30,8 +30,11 @@ auto AtomBase::size() const -> size_t {
 }
 
 auto AtomBase::at(size_t index) const -> value_type {
+    /*
```
BenKaufmann: This change looks weird.

BenKaufmann:
> @BenKaufmann just for your info: here is the work in progress to get things running on windows

Interesting. I also had to define `ssize_t` in various places.

rkaminsk:
> > @BenKaufmann just for your info: here is the work in progress to get things running on windows
> 
> Interesting. I also had to define `ssize_t` in various places.

I just used ptrdiff_t because I think it should match size_t.

lib/c-api/src/ast_yaml.cc
rkaminsk: Maybe I'll do that. The string is just used for code generation. So it is not terrible important to have on Windows.

rkaminsk:
I still don't get why the aspif tests fail. I suspect something about the temporary files created in these tests. In my vm, the tests run through now.

Issue #28, opened 2017-03-30
confusing result of weak constraint


wangbiu:
results of programs with weak constraints confuse me, for example:  
```
a | b .
:~ a. [2@1]
:~ b. [1@1]
```
according to semantics of weak constraints, the optimal answer sets should be `{b}`, but output of clingo is 
![lwoc j 5 9 o zwg_4](https://cloud.githubusercontent.com/assets/7279980/24504204/e35d9d82-1587-11e7-8a15-e3bf344ec0be.png)

the command line I used is `clingo file`

Aluriak:
Solutions are printed in the order they are found by the solver. Consequently, the optimal answer is the last answer, here the number 2.
The lines

    Answer: 2
    b
    Optimization: 1
    OPTIMUM FOUND

describes that the answer set number 2 contains atom `b`, reach an optimization of 1 and is an optimum.

rkaminsk:
You might want to use option `-q1` to only print the optimal solution. Another interesting option is `--opt-mode=optN` to enumerate all optimal solutions. I am going to tag this as invalid and close the issue. -R

PR #555, opened 2025-07-10
Issue #28: Unify PropagateControl and PropagateInit

BenKaufmann:
* update clasp
* adjust C-API:
  - drop clingo_propagate_init_add_watch{_to_thread}
  - drop clingo_propagate_init_remove_watch{_from_thread}
  - drop clingo_propagate_init_assignment
  - drop clingo_propagate_init_add_literal
  - drop clingo_propagate_init_add_clause
  - drop clingo_propagate_init_propagate
  - drop clingo_propagate_control_thread_id
  - drop clingo_propagate_control_assignment
  - add clingo_propagate_init_control for converting a
    clingo_propagate_init_t into a clingo_propagate_control_t object
  - adjust signatures of clingo_propagator functions: thread_id and
    current assignment are now explicit function parameters
  - **add clingo_propagator_attach_callback_t for initializing solving
    threads**
* adjust C++-API and tests
* adjust Python-API and tests

### DIFF
--- a/lib/c-api/include/clingo/propagate.h
+++ b/lib/c-api/include/clingo/propagate.h
@@ -56,6 +56,14 @@ typedef int clingo_truth_value_t;
 
 //! @name Assignment Functions
 //! @{
+//! Get the id of the thread to which the assignment belongs.
+//!
+//! @note Thread ids are consecutive numbers starting with zero.
+//!
+//! @param[in] assignment the target assignment
+//! @param[out] id the thread id
+//! @return whether the call was successful
+CLINGO_VISIBILITY_DEFAULT bool clingo_assignment_thread_id(clingo_assignment_t const *assignment, clingo_id_t *id);
 
 //! Get the current decision level.
 //!
@@ -229,6 +237,9 @@ enum clingo_weight_constraint_type_e {
 //! Corresponding type to ::clingo_weight_constraint_type_e.
 typedef int clingo_weight_constraint_type_t;
 
+//! This object can be used to add clauses and propagate literals while solving.
+typedef struct clingo_propagate_control clingo_propagate_control_t;
+
 //! Object to initialize a user-defined propagator before each solving step.
 //!
 //! Each @link c_base symbolic or theory atom@endlink is uniquely associated with an
@@ -254,38 +265,7 @@ typedef struct clingo_propagate_init clingo_propagate_init_t;
 CLINGO_VISIBILITY_DEFAULT bool clingo_propagate_init_solver_literal(clingo_propagate_init_t const *init,
                                                                     clingo_literal_t aspif_literal,
                                                                     clingo_literal_t *solver_literal);
-//! Add a watch for the solver literal in the given phase.
-//!
-//! @param[in] init the target
-//! @param[in] solver_literal the solver literal
-//! @return whether the call was successful
-CLINGO_VISIBILITY_DEFAULT bool clingo_propagate_init_add_watch(clingo_propagate_init_t *init,
-                                                               clingo_literal_t solver_literal);
-//! Add a watch for the solver literal in the given phase to the given solver thread.
-//!
-//! @param[in] init the target
-//! @param[in] solver_literal the solver literal
-//! @param[in] thread_id the id of the solver thread
-//! @return whether the call was successful
-CLINGO_VISIBILITY_DEFAULT bool clingo_propagate_init_add_watch_to_thread(clingo_propagate_init_t *init,
-                                                                         clingo_literal_t solver_literal,
-                                                                         clingo_id_t thread_id);
-//! Remove the watch for the solver literal in the given phase.
-//!
-//! @param[in] init the target
-//! @param[in] solver_literal the solver literal
-//! @return whether the call was successful
-CLINGO_VISIBILITY_DEFAULT bool clingo_propagate_init_remove_watch(clingo_propagate_init_t *init,
-                                                                  clingo_literal_t solver_literal);
-//! Remove the watch for the solver literal in the given phase from the given solver thread.
-//!
-//! @param[in] init the target
-//! @param[in] solver_literal the solver literal
-//! @param[in] thread_id the id of the solver thread
-//! @return whether the call was successful
-CLINGO_VISIBILITY_DEFAULT bool clingo_propagate_init_remove_watch_from_thread(clingo_propagate_init_t *init,
-                                                                              clingo_literal_t solver_literal,
-                                                                              uint32_t thread_id);
+
 //! Freeze the given solver literal.
 //!
 //! Any solver literal that is not frozen is subject to simplification and might be removed in a preprocessing step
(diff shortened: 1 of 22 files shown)


BenKaufmann:
@rkaminsk Marked as draft for now since I'd like to get your feedback first. E.g., do you think `PropagateControl` is still a good name? Do you think the (python) API changes (in particular regarding the Propagator) are too disrupting?

rkaminsk:
> @rkaminsk Marked as draft for now since I'd like to get your feedback first. E.g., do you think `PropagateControl` is still a good name? Do you think the (python) API changes (in particular regarding the Propagator) are too disrupting?

Thank's I'll try to have a closer look asap. Regarding the API changes, yes I think this is fine. Since we go from clingo 5 to 6, now is the chance. Regarding the name PropagateControl, I don't have a better idea. The language model was suggesting PropgationControllor or PropagationContext but that seems very similar to me.

BenKaufmann:
> Regarding the name  PropagateControl, I don't have a better idea. The language model was suggesting PropgationControllor or PropagationContext but that seems very similar to me.

To me, all of these strongly hint at "propagation" but with this PR, the control is no longer limited to propagation.

I thought about "PropagatorControl" but that's not great either 🤷‍♂️

lib/c-api/include/clingo/propagate.h
```diff
@@ -530,25 +454,31 @@ CLINGO_VISIBILITY_DEFAULT bool clingo_propagate_control_add_clause(clingo_propag
...
+//! @return whether the call was successful
 CLINGO_VISIBILITY_DEFAULT bool clingo_propagate_control_propagate(clingo_propagate_control_t *control, bool *result);
 
 //! @}
 
 //! Typedef for @ref ::clingo_propagator::init().
-typedef bool (*clingo_propagator_init_callback_t)(clingo_propagate_init_t *, void *);
+typedef bool (*clingo_propagator_init_callback_t)(clingo_assignment_t const *, clingo_propagate_init_t *, void *);
+
+//! Typedef for @ref ::clingo_propagator::attach().
```
BenKaufmann, line 475: @rkaminsk We could consider making the `thread_id` a (read-only) property of the `Assignment`. This would reduce the number of parameters in the callbacks, and it is still conceptually sound - in the context of `init()`, the assignment belongs to thread 0.

lib/c-api/include/clingo/propagate.h
```diff
@@ -402,7 +351,7 @@ CLINGO_VISIBILITY_DEFAULT bool clingo_propagate_init_add_clause(clingo_propagate
 //! @param[in] type the type of the weight constraint
 //! @param[in] compare_equal if true compare equal instead of less than equal
 //! @param[out] result result indicating whether the problem became unsatisfiable
-//! @return wether the call was successful
+//! @return whether the call was successful
```
BenKaufmann: @rkaminsk In contrast to `clingo_propagate_init_add_minimize`, we could move `clingo_propagate_init_add_weight_constraint` to the propagate control  interface thereby making it available also to individual solver threads (during propagation/checking). Do you see any need for this?

Similarly for `clingo_propagate_init_number_of_threads`.

rkaminsk:
> > Regarding the name  PropagateControl, I don't have a better idea. The language model was suggesting PropgationControllor or PropagationContext but that seems very similar to me.
> 
> To me, all of these strongly hint at "propagation" but with this PR, the control is no longer limited to propagation.
> 
> I thought about "PropagatorControl" but that's not great either 🤷‍♂️

Then, let's just stick with PropagateControl. The purpose of the class goes a bit beyond simple propagation in the sense that we can also add problem clauses. But controlling propagation it is still one of its main tasks.

lib/c-api/include/clingo/propagate.h
rkaminsk: This would allow us to play with pb cuts. 👍for the threads.

lib/c-api/include/clingo/propagate.h
BenKaufmann: Ok. Will do so.

lib/c-api/include/clingo/propagate.h
BenKaufmann: > 👍for the threads.

Ok. So `clingo_propagate_init_number_of_threads` will become `clingo_propagate_control_number_of_threads`.

> This would allow us to play with pb cuts.

Should I add it to this PR, then? If so, do we need the `tag` capability?

rkaminsk: > `clingo_assignment_thead_id`

This is a real improvement to the interface we are building.

> `clingo_propagate_control_number_of_threads`

Now that I think about it, maybe we should leave this in the init interface. Logically, it makes most sense there. A propagator can reserve space based on this value. During propagation, it should never be necessary to use it (and in the unlikely case that it is, it can still be stored as a member).

> `clingo_propagate_init_add_weight_constraint`

If it is not too much work to get this implemented, this would be nice to have. Maybe even a nice student topic to build something on top of this.

BenKaufmann: > > `clingo_propagate_control_number_of_threads`
> 
> Now that I think about it, maybe we should leave this in the init interface. 

Ups. Going to revert the change then :smile: 

> > `clingo_propagate_init_add_weight_constraint`
> 
> If it is not too much work to get this implemented, this would be nice to have. 

The amount of work necessary depends on our requirements.  If we don't need tagging (step-local weight constraints) and dynamic deletion of weight constraints, then implementation is quite trivial. Additionally adding tagging also shouldn't be too problematic. However, properly integrating weight constraints into the various deletion heuristics would require quite some effort.

rkaminsk: > > > `clingo_propagate_control_number_of_threads`
> > 
> > 
> > Now that I think about it, maybe we should leave this in the init interface.
> 
> Ups. Going to revert the change then 😄

This is probably not a big thing anyway.

> 
> > > `clingo_propagate_init_add_weight_constraint`
> > 
> > 
> > If it is not too much work to get this implemented, this would be nice to have.
> 
> The amount of work necessary depends on our requirements. If we don't need tagging (step-local weight constraints) and dynamic deletion of weight constraints, then implementation is quite trivial. Additionally adding tagging also shouldn't be too problematic. However, properly integrating weight constraints into the various deletion heuristics would require quite some effort.

Then maybe we should restrict to a simple implementation first, mark the feature as experimental, and continue to refine the implementation if there is an actual need...

BenKaufmann:
@rkaminsk This should be ready now.  Weight constraints can now be added via the control object.
I prepared a clingcon branch with (minimal) required changes here: https://github.com/potassco/clingcon/tree/issue-28

One thing still missing is propagation support in the SAT-preprocessor but this is a clasp-only change.

rkaminsk:
> @rkaminsk This should be ready now. Weight constraints can now be added via the control object. I prepared a clingcon branch with (minimal) required changes here: https://github.com/potassco/clingcon/tree/issue-28
> 
> One thing still missing is propagation support in the SAT-preprocessor but this is a clasp-only change.

Thanks, for the effort. I'll have a look and merge next week.

PR #559, opened 2025-07-31
extend config api to allow for adding custom entries

### LINKED ISSUE #562: Bug generating docstrings for ast's (wip-20)
Parameters with multiline docstrings are not correctly handled.

rkaminsk:
This PR extends the API to allow for adding entries to the Config of the Control object.

@BenKaufmann, what do you think about the API? Any suggestions?

## TODO

- [x] C API
- [x] Python API
  - [x] add docstrings for callbacks
  - [x] add tests
  - [x] proper lifetime management for config entries
- [x] C++ API
  - [x] add tests
- [ ] polish
- [ ] think about integrating config and application options

## Example

```python

lib = Library()
ctl = Control(lib)
cfg = ctl.config


class PluginConfig:
    def __init__(self):
        self.value = "default value"
        self.array = ["default value"]
        self.group = "some group"

    def set_val(self, value, index=None):
        self.value = value

    def get_val(self, index=None):
        return self.value

    def set_arr_len(self, value, index=None):
        while len(self.array) < int(value):
            self.array.append("default value")

    def get_arr_len(self):
        return len(self.array)

    def set_arr_val(self, value, index=None):
        if index is None:
            index = 0
        self.array[index] = value

    def get_arr_val(self, index=None):
        if index is None or index >= len(self.array):
            return "undefined"
        return self.array[index]

    def get_group(self, index=None):
        return self.group

    def set_group(self, value, index=None):
        self.group = value


pc = PluginConfig()

cfg.add_entry("plugin", "a plugin that needs configuration")
cfg.add_entry("plugin.group", "a group", pc.get_group, pc.set_group)
cfg.add_entry("plugin.group.val", "a value", pc.get_val, pc.set_val)
cfg.add_entry("plugin.array[]", "an array", set=pc.set_arr_len, size=pc.get_arr_len)
cfg.add_entry("plugin.array.val", "an array value", pc.get_arr_val, pc.set_arr_val)

assert str(cfg.plugin.group.val.value) == "default value"
cfg.plugin.group.val = "new value"
assert str(cfg.plugin.group.value) == "some group"
assert str(cfg.plugin.group.val) == '"new value"'
assert cfg.plugin.group.val.value == "new value"
cfg.plugin.group.value = "test"
assert cfg.plugin.group.value == "test"

assert len(cfg.plugin.array) == 1
cfg.plugin.array.value = "2"
assert len(cfg.plugin.array) == 2

assert cfg.plugin.array[0].val.value == "default value"
assert cfg.plugin.array.val.value == "undefined"
cfg.plugin.array[1].val = "5"

print(cfg.plugin)
```


lib/control/include/clingo/control/config.hh
```diff
@@ -0,0 +1,366 @@
+#pragma once
+
+#include <clingo/util/enum.hh>
+#include <clingo/util/print.hh>
+
+#include <clasp/cli/clasp_options.h>
+
+#include <map>
+
```
BenKaufmann, line 10: Consider including `<sstream>` , which is needed for `std::ostringstream` used in `error_()`

rkaminsk: Thanks, what do you think about the interface. I certainly will make  another pass over the code as such.

lib/control/include/clingo/control/config.hh
```diff
@@ -0,0 +1,366 @@
...
+    auto parse_path_(Key key, std::string_view path) const -> std::optional<Key>;
+    // Checks if this key is a clasp key.
+    //
+    // Returns true if the upper bit is not set.
+    [[nodiscard]] static auto is_clasp(KeyType key) -> bool;
+
+    // Checks if this key is a clingo key.
+    //
+    // Returns true if the upper bit is set and the key is not the invalid key.
+    [[nodiscard]] static auto is_clingo(KeyType key) -> std::optional<Key>;
```
BenKaufmann, line 360: The documentation is misleading as the function does not return bool.
Furthermore, from what I can see `is_clingo(key_invalid())` does not return a `nullopt` but would throw an exception.

lib/control/include/clingo/control/config.hh
```diff
@@ -0,0 +1,366 @@
...
+
+    // Get the string representation of the configuration tree under the given key.
+    void str_(Util::OutputBuffer &out, KeyType key, size_t first_indent, size_t indent) const;
+
+    Clasp::Cli::ClaspCliConfig *config_;
+    std::string mutable buf_;
```
BenKaufmann, line 366: Is it guaranteed that objects of this class are never accessed from multiple threads?

EDIT: Nevermind. Missed the corresponding note in the class description.

lib/control/include/clingo/control/config.hh
```diff
@@ -0,0 +1,366 @@
...
+    void set_value(KeyType key, std::string_view value);
+
+    //!  Adds a new clingo configuration entry to the configuration tree.
+    //!
+    //!  Validates the key name format and inserts the key as a child of the
+    //!  given parent key. The key name may include an optional index in the form
+    //!  "key[index]" to add an array key. The index of such a key is used as
+    //!  default index when the key is accessed without an index.
+    //!
+    //!  If an updater is provided, it will be associated with the key to handle
```
BenKaufmann, line 176: What is an _updater_?

lib/control/include/clingo/control/config.hh
```diff
@@ -0,0 +1,366 @@
...
+        [[nodiscard]] auto key_id() const -> KeyType;
+        //! Get the index of the key.
+        [[nodiscard]] auto index() const -> std::optional<KeyType>;
+        //! Transfer the array index of this key to the child key.
+        [[nodiscard]] auto subkey(Key child) const -> Key;
+        //! Get the raw representation of the key.
+        [[nodiscard]] auto rep() const -> KeyType;
+
+        //! Compare two keys for equality.
+        [[nodiscard, maybe_unused]] friend auto operator==(Key const &a, Key const &b) -> bool {
```
BenKaufmann: Consider using `=default` here and below.

lib/control/include/clingo/control/config.hh
rkaminsk: Yes, docs outdated but the behavior is intended like this.

lib/control/include/clingo/control/config.hh
rkaminsk: Imo it just makes no sense to provide a thread-safe config. I won't bother to support this.

lib/control/include/clingo/control/config.hh
rkaminsk: First it was an updater and then it became the entry interface.

lib/c-api/include/clingo/config.h
```diff
@@ -42,9 +42,92 @@ enum clingo_config_type_e {
...
+//! @param[in] config the configuration object
+//! @param[in] parent the parent key under which to add the entry
+//! @param[in] name the name of the new entry (not null-terminated)
+//! @param[in] name_size the length of the name string
+//! @param[in] description the description of the entry (not null-terminated)
+//! @param[in] description_size the length of the description string
+//! @param[in] entry pointer to the entry callbacks struct
+//! @param[in] data user data pointer for the entry callbacks
+//! @return true on success, false on error
+CLINGO_VISIBILITY_DEFAULT bool clingo_config_add(clingo_config_t *config, clingo_id_t parent, char const *name,
```
BenKaufmann, line 110: It is consistent with other API functions and you probably already explained this to me but could you briefly reiterate why we pass "string views" as two parameters but receive them as a `clingo_string_t`?

BenKaufmann:
> @BenKaufmann, what do you think about the API? 

The API looks good to me. However, I'm currently wondering about the connection between config entries and options.
Is there such a connection and if not, do you think it would make sense to somehow (automatically) add new options to the configuration?

lib/c-api/include/clingo/config.h
```diff
@@ -138,9 +209,10 @@ CLINGO_VISIBILITY_DEFAULT bool clingo_config_map_subkey_name(clingo_config_t con
 //! @param[in] name the name to look up the subkey
 //! @param[in] size the size of the name
 //! @param[out] subkey the resulting subkey
+//! @param[out] has_subkey whether the map has the subkey
 //! @return whether the call was successful
 CLINGO_VISIBILITY_DEFAULT bool clingo_config_map_at(clingo_config_t const *config, clingo_id_t key, char const *name,
-                                                    size_t size, clingo_id_t *subkey);
+                                                    size_t size, clingo_id_t *subkey, bool *has_subkey);
```
BenKaufmann, line 198: What's the motivation for this additional out-parameter? Couldn't we return `false` (and set clingo_result_range) if `name` is not an element of the map?

lib/control/include/clingo/control/config.hh
```diff
@@ -0,0 +1,365 @@
...
+
+    //! Retrieves information about the specified key.
+    //!
+    //! @note All out parameters are optional (i.e., can be null).
+    //!
+    //! @param key a handle to a key
+    //! @param n_children the number of subkeys for this key
+    //! @param array_info the length of array keys or -1 for non-array keys
+    //! @param value_info whether the key has a value, can be read, and can be set
+    //! @param help a description of the key
```
BenKaufmann: `help` is not a parameter

lib/control/include/clingo/control/config.hh
```diff
@@ -0,0 +1,365 @@
...
+    //!
+    //! @param key key of the array entry
+    //! @param index index of the array element
+    //! @return key for the array element at the given index
+    [[nodiscard]] auto array_at(KeyType key, KeyType index) const -> KeyType;
+
+    //! Get the key for a named subkey in a map entry.
+    //!
+    //! Returns the key corresponding to the subkey with the given path in the
+    //! map configuration entry specified by `key`. Throws if the subkey does
```
BenKaufmann: Is this true and if the function indeed throws if the subkey does not exist, couldn't we return `KeyType` then?

lib/control/include/clingo/control/config.hh
```diff
@@ -0,0 +1,365 @@
...
+    //!
+    //! @param key key of the map entry
+    //! @param index index of the subkey
+    //! @return name of the nth subkey
+    [[nodiscard]] auto map_nth(KeyType key, KeyType index) const -> std::string_view;
+
+    //! Get the value of a configuration entry.
+    //!
+    //! Returns the value associated with the given configuration key.
+    //! Throws if the key does not have a value or cannot be accessed.
```
BenKaufmann: It seems that the implementation actually distinguishes between "does not have a value" and "value cannot be accessed".

rkaminsk: Indeed it does. It throws if there is no getter and returns nullopt if there is a getter but no value. At least that is the idea I had.

lib/control/include/clingo/control/config.hh
rkaminsk: I'll refine. The optional is needed for the C API. The initial version did throw but then this changed. 😅

lib/c-api/include/clingo/config.h
rkaminsk: The idea of clingo's errors is to indicate that something happened that shouldn't have happened. The out parameter is there to implement something like a contains.

lib/c-api/include/clingo/config.h
rkaminsk: I use string for output and char + size for input.

rkaminsk:
> > @BenKaufmann, what do you think about the API?
> 
> The API looks good to me. However, I'm currently wondering about the connection between config entries and options. Is there such a connection and if not, do you think it would make sense to somehow (automatically) add new options to the configuration?

There is no connection atm. I was thinking to keep it separate to make my life easier. In the current API it would be difficult to bring the two together. I think...

lib/c-api/include/clingo/config.h
rkaminsk: Also doing the input would prob be good. Its just a pain to change.

BenKaufmann: > Also doing the input would prob be good. Its just a pain to change.

...but you have AI now to alleviate the pain :sweat_smile:

lib/cxx-api/tests/config.cc
rkaminsk, line 1: @BenKaufmann, what do you think about this C++ interface to extend the config? The tests in this commit show how it can be used.

lib/cxx-api/include/clingo/config.hh
```diff
@@ -179,6 +179,86 @@ class Config : public ConstConfig {
...
+        static constexpr auto has_set = requires { std::declval<EntryType>().set(std::string_view{}); };
+        static constexpr auto has_array_set =
+            requires { std::declval<EntryType>().set(std::optional<size_t>{}, std::string_view{}); };
+        static constexpr auto is_array = requires { entry.size(); };
+
+        static_assert(!has_get || !has_array_get, "entry cannot have both get with and without index");
+        static_assert(!has_set || !has_array_set, "entry cannot have both set with and without index");
+
+        static constexpr auto c_entry = clingo_config_entry_t{
+            has_get || has_array_get ? *[](size_t const *index, void *data, clingo_string_t *value, bool *has_value) -> bool {
```
BenKaufmann: Nitpick: I think using unary `+` to force the conversion to a function pointer is a slightly more common (see e.g. https://medium.com/@simontoth/daily-bit-e-of-c-lambda-to-function-pointer-b2aef7507b70)

lib/cxx-api/include/clingo/config.hh
```diff
@@ -179,6 +179,86 @@ class Config : public ConstConfig {
...
+                        *has_size = is_array;
+                    }
+                }
+                CLINGO_CATCH;
+            } : nullptr,
+            [](void *data) { std::unique_ptr<EntryType> self{static_cast<EntryType *>(data)}; },
+        };
+
+        auto copy = std::make_unique<EntryType>(std::forward<Entry>(entry));
+        Detail::handle_error(clingo_config_add(cfg_(), key_, name.data(), name.size(), description.data(),
```
BenKaufmann, line 278: Hm, isn't this a memory leak if `clingo_config_add` returns an error?
I.e. shouldn't this be something like:
```cpp
auto copy = std::make_unique<EntryType>(std::forward<Entry>(entry));
Detail::handle_error(clingo_config_add(cfg_(), key_, name.data(), name.size(), description.data(),
                                               description.size(), &c_entry, copy.get()));
copy.release();
```

lib/cxx-api/tests/config.cc
```diff
@@ -73,4 +73,63 @@ TEST_CASE_METHOD(Fixture, "config solve", "[cxx][config][solve]") {
...
+        [[nodiscard]] auto size() const -> size_t { return values_->size(); }
+
+      private:
+        std::vector<std::optional<std::string>> *values_;
+    };
+    class ConfigArrayValue {
+      public:
+        ConfigArrayValue(std::vector<std::optional<std::string>> &values) : values_{&values} {}
+        [[nodiscard]] auto get(std::optional<size_t> index) const -> std::optional<std::string> {
+            return values_->at(index ? *index : 0);
```
BenKaufmann: Nitpick: Consider using `value_or` here and below.

lib/cxx-api/tests/config.cc
BenKaufmann: It took me a while to understand the difference between `ConfigArray` and `ConfigArrayValue` but otherwise, looks ok to.

Out of curiosity: what's the motivation for not allowing set/get for both with and without index?

lib/cxx-api/include/clingo/config.hh
rkaminsk: I have to check the implementation of `clingo_config_add`. I usually try to make sure the C function guarantees cleanup. It seems like I might have missed it here though.

lib/cxx-api/tests/config.cc
rkaminsk: The idea is that either an entry is in an array or it is not. Then it would not make sense to have both. This might catch some c&p errors early on.

lib/c-api/include/clingo/config.h
rkaminsk: I remember that some ffis also had problems passing structs by value, which would be the natural thing to do with input strings. I'll just keep it as is. It seems just a minor inconvenience.

lib/cxx-api/include/clingo/config.hh
rkaminsk: Thanks, fixed.

lib/cxx-api/include/clingo/config.hh
rkaminsk: Fixed.

lib/cxx-api/tests/config.cc
rkaminsk: Changed.

PR #560, opened 2025-08-07
Minor profile cleanup

BenKaufmann:
### fix potential UB in end_step()
* end_step() dereferenced the passed in grounder pointer in its
  internal loop without checking for nullptr first. Change signature
  so that it now takes the grounder by reference.

### simplify profile output handling
* move step profile output logic to Grounder::print_summary()
* add event type for signaling that grounding has finished and
  let ClingoApp handle it by calling Grounder::print_summary()
* NOTE: for consistency with the old behavior, the event is not
  emitted when grounding is invoked from a backend handler -
  in that case, the params vector is always empty and no profile
  data is created

### DIFF
--- a/lib/control/include/clingo/control/solver.hh
+++ b/lib/control/include/clingo/control/solver.hh
@@ -582,7 +582,7 @@ class Grounded : public Clasp::Event {
   public:
     explicit Grounded(Input::ProgramParamVec const &params)
         : Event(this, subsystem_load, verbosity_quiet), params(params) {}
-    Input::ProgramParamVec const &params; // NOLINT
+    std::span<Input::ProgramParamVec::value_type const> params;
 };
 
 //! A grounder and solver for logic programs.
--- a/lib/control/include/clingo/control/solver.hh
+++ b/lib/control/include/clingo/control/solver.hh
@@ -576,6 +576,15 @@ class SymbolTable {
 //! A unique pointer to a symbol table.
 using USymbolTable = std::unique_ptr<SymbolTable>;
 
+//! Event emitted by a solver after the program is grounded.
+// TODO: Simplify/Move on next output refactoring
+class Grounded : public Clasp::Event {
+  public:
+    explicit Grounded(Input::ProgramParamVec const &params)
+        : Event(this, subsystem_load, verbosity_quiet), params(params) {}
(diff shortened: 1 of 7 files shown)


lib/control/tests/text.cc
```diff
@@ -16,7 +16,7 @@ TEST_CASE("grounder_text") {
         log.set_level(LogLevel::error);
         auto buf = Util::OutputBuffer{};
         auto out = Output::make_text_output(buf);
-        Control::Grounder grd{log, *store, opts, *out, false};
+        Control::Grounder grd{log, *store, opts, *out};
```
rkaminsk:
```suggestion
        auto grd = Control::Grounder{log, *store, opts, *out};
```

lib/control/include/clingo/control/solver.hh
```diff
@@ -576,6 +576,15 @@ class SymbolTable {
 //! A unique pointer to a symbol table.
 using USymbolTable = std::unique_ptr<SymbolTable>;
 
+//! Event emitted by a solver after the program is grounded.
+// TODO: Simplify/Move on next output refactoring
+class Grounded : public Clasp::Event {
+  public:
+    explicit Grounded(Input::ProgramParamVec const &params)
+        : Event(this, subsystem_load, verbosity_quiet), params(params) {}
+    Input::ProgramParamVec const &params; // NOLINT
```
rkaminsk: I was consistently using pointers + maybe a getter to never store any refs. But maybe it is best to use a span here and just pass by value.

rkaminsk (approved):
Thanks, looks better than before. Just two nitpicks: one in the old and in the new code.

rkaminsk:
> * NOTE: for consistency with the old behavior, the event is not
>       emitted when grounding is invoked from a backend handler -
>       in that case, the params vector is always empty and no profile
>       data is created

This makes sense. Ground is called with an empty parameter list to put Control::Solver into the right state. There is no need to output profiling data in this case.

lib/control/tests/text.cc
BenKaufmann: Done.

lib/control/include/clingo/control/solver.hh
BenKaufmann: Switched to std::span.

PR #580, opened 2025-11-05
add function for asynchronous grounding

### LINKED ISSUE #570: Making grounding interruptable at a timeout boundary
I'd like to know if Potassco would be willing to accept a pull request that plumbs a timeout down into the grounder's queue processing.

### The problem

We have a modeling system that could be described in the abstract as "Try lots of programs for the same problem, see which ones work".

One way a particular model candidate can "not work" is if it takes longer than a few seconds to run.

When we have code like this:

```python
def run_clingo(asp_string: str) -> tuple[list, str]:
    ctrl = Control(logger=clingo_logger)
    ctrl.add("base", [], asp_string)
    ctrl.ground([("base", [])])
    model_lines = []

    def on_model(m):
        model_lines.append([str(sym) for sym in m.symbols(shown=True)])

    result = "UNKNOWN"
    solve_timeout_in_seconds = 5
    with ctrl.solve(on_model=on_model, async_=True) as handle:
        finished = handle.wait(solve_timeout_in_seconds)
        if not finished:
            logger.warning(f"ASP solve timed out after {solve_timeout_in_seconds} seconds")
        else:
            solve_result = handle.get()
            if solve_result.satisfiable:
                result = "SATISFIABLE"
            else:
                result = "UNSATISFIABLE"

    return model_lines, result
```

We can successfully abort if the SOLVER is taking to long, but if the program introduces some kind of large domain that is going to take a while to fully ground, we have not found a way to interrupt it without killing the process.  

### The current mitigation

What we do today is actually boot up a sidecar python process to be the "clingo server" and use local sockets as IPC to enqueue models to try; this works fairly well, but does include the overhead of having to communicate between the two processes, and during a timeout scenario we have to kill that server process and boot a new one (which also takes time) and recover the queue state for any in flight work.  

It's workable, but many of our developers don't have a great mental model about why we use that pattern, and it's not ideal for packing as many worker processes as possible onto a node at once.

### What I would like to do

If you'd accept the PR (or at least be willing to review and consider it), I'd like to work on plumbing an optional timeout parameter through the "ground" operation.  Something like:

```python
 ctrl.ground([("base", [])], timeout_in_seconds: 3)
```

Of course I'd use the API pattern of your choice, but the main idea is to provide some channel down to the gringo queue processing so that every N messages, if the provided timeout value is non-zero, it would check the system clock and throw an error if the timeout had elapsed.

### Backwards compatibility

I think this might be safe to do because any code not providing this parameter (which would be all existing clingo using code) would not provide that parameter, and thus would never receive the exception. Code using the new parameter could explicitly try/except in the python process itself and not have to run a different process to enforce a timeout.

What do y'all think?  Would you be willing to review and consider such a change?

Alternatively, is there some established pattern for managing this (within a single python process) that I'm not currently aware of?

Thanks for your time,

~Ethan

rkaminsk:
This PR adds asynchronous grounding capabilities to the internal C++ API and refactors the solve API for consistency. The `solve` function now executes synchronously with callbacks, while a new `start_solve` function provides the previous async behavior. Similarly, a new `start_ground` function enables background grounding with control via a ground handle.

Key changes:
- Split `solve` into synchronous `solve` and asynchronous `start_solve` functions
- Added `start_ground` for background grounding operations
- Updated all test files to use the new API patterns
- Converted Python enums to native IntEnum for better type compatibility

Per file summary:

| File | Description |
| ---- | ----------- |
| lib/util/include/clingo/util/sync.hh | Added StopFlag class for signaling grounding interruption |
| lib/python-api/src/ground.hh | New header defining GroundHandle for async grounding |
| lib/python-api/src/ground.cc | Implementation of ground handle and result types |
| lib/python-api/src/control.cc | Added start_ground/start_solve, refactored solve API |
| lib/python-api/src/solve.cc | Updated solve handle closing to not throw |
| lib/python-api/src/util.hh | Fixed GC traversal bug for unconstructed objects |
| lib/ground/src/instantiator.cc | Added stop flag support for grounding interruption |
| scripts/stubs.py | Updated enum generation for native enums |
| scripts/ast_module.j2 | Switched to native_enum for enum bindings |
| lib/python-api/tests/*.py | Updated tests to use new solve/start_solve API |
| lib/cxx-api/tests/*.cc | Updated C++ tests to use new API patterns |
| third_party/* | Updated submodule commits |


lib/util/include/clingo/util/sync.hh
```diff
@@ -0,0 +1,26 @@
...
+    StopFlag(StopFlag const &other) = delete;
+    StopFlag(StopFlag &&other) noexcept = delete;
+    auto operator=(StopFlag const &other) -> StopFlag & = delete;
+    auto operator=(StopFlag &&other) noexcept -> StopFlag & = delete;
+
+    void request_stop() { state_.store(true, std::memory_order_release); }
+    auto stop_requested() -> bool { return state_.load(std::memory_order_acquire); }
+
+  private:
+    std::atomic<bool> state_ = false;
```
BenKaufmann, line 30: You could use a `std::atomic_flag` here, which already has the special members deleted.

lib/control/include/clingo/control/grounder.hh
```diff
@@ -12,6 +12,63 @@ namespace CppClingo::Control {
...
+    //!
+    //! The function returns true if grounding has finished, false otherwise.
+    //!
+    //! @param timeout The maximum time to wait in seconds.
+    [[nodiscard]] auto wait(double timeout) -> bool;
+    //! Get the result of grounding.
+    //!
+    //! Blocks until grounding is finished. Raises exceptions thrown during
+    //! grounding.
+    [[nodiscard]] auto get() -> GroundResult;
```
BenKaufmann: Hm. This API looks rather dangerous to me given that:
- calling `std::future::get()` on a future for which `valid()` is false results in UB
- calling `std::future::get()` turns a `valid()` future into future for which `valid()` is false
- GroundHandle has no API for testing whether the underlying future is valid.

Furthermore, I would consider making `cancel()` noexcept. Instead, I would allow a call to `get()` after `cancel()`. This way, clients can decide whether they care about the result of a cancelled operation and additionally  `op=` could then also become noexcept.

rkaminsk: As far as I can see, there is no noexcept waiting. So a destructor/cancel statically guaranteed without exceptions. Proper API-usage might still make them impossible. And you are right, we can only obtain the result of a future once. I'll think a bit more about this.

lib/control/src/grounder.cc
```diff
@@ -369,7 +372,89 @@ void Grounder::accept(Ground::ProfileNode::Visitor const &visit) const {
...
+    Impl(Impl &&other) noexcept = delete;
+    auto operator=(Impl const &) -> Impl & = delete;
+    auto operator=(Impl &&other) noexcept -> Impl & = delete;
+
+    void start(Grounder &grd, Input::ProgramParamVec params, Ground::ScriptCallback *context) {
+        assert(!future_.valid());
+        auto task = std::packaged_task<GroundResult()>{
+            [this, &grd, params = std::move(params), context]() { return grd.ground(params, context, &stop_); }};
+        future_ = task.get_future();
+        thread_ = std::thread{std::move(task)};
```
BenKaufmann: AFAICS, you could get rid of the explicit thread and the packaged_task and just use `std::async` with `launch::async`.

The dtor would then basically just need to call `cancel()`.

rkaminsk: I'll have a look.

rkaminsk: This made it quite a bit easier.

lib/util/include/clingo/util/sync.hh
rkaminsk: I find it's API a bit strange for the use case here. I'll have to use test_and_set instead of a simple store.

lib/cxx-api/include/clingo/control.hh
```diff
@@ -349,6 +435,63 @@ class Control {
...
+            parts = default_parts;
+        }
+        if (parts) {
+            c_parts.reserve(parts->size());
+            for (auto const &part : *parts) {
+                c_parts.emplace_back(part.name.data(), part.name.size(), c_cast(part.params.data()),
+                                     part.params.size());
+            }
+        } else {
+            c_parts.reserve(1);
```
BenKaufmann: What's the motivation for this `reserve()` call? AFAIK most stdlibs only allocate 1 slot on first push anyway. Furthermore, `reserve()` is also allowed to "over"-allocate.

rkaminsk: Just dutifully following reserve known size + insert afterward. I guess it does not hurt but also does not really help. At least for the libc++ I quickly checked and it is just using __builtin_operator_new without any magic.

rkaminsk: I would also find an STL where the *initial* reserve overallocates rather strange. I can't imagine anyone doing this. I agree however that the reserve is unnecessary for a single element.

lib/cxx-api/include/clingo/control.hh
```diff
@@ -349,6 +454,92 @@ class Control {
...
+                                                         c_data != nullptr ? &c_handler : nullptr, c_data, &handle));
+        return GroundHandle{handle};
+    }
+
+    //! @copydoc start_ground
+    template <GroundEventHandlerLike Handler = std::nullptr_t>
+    [[nodiscard]] auto start_ground(std::initializer_list<Part> parts, Handler &&handler = nullptr) const
+        -> GroundHandle {
+        return start_ground(PartSpan{parts}, std::forward<Handler>(handler));
+    }
```
rkaminsk, line 353: @BenKaufmann what do you think about this implementation of start_ground. Specially, the way the lifetime of the handler is managed.

BenKaufmann: I think a slightly more idiomatic way would be to always pass handler by value, i.e. `Handler handler` and then let the caller use `std::ref(myHandler)` to signal "pass-by-value".

Your function would then become one of the few functions with special handling for `std::reference_wrapper`.

lib/cxx-api/include/clingo/control.hh
```diff
@@ -349,6 +454,92 @@ class Control {
...
+                }
+                CLINGO_CATCH;
+            },
+            [](clingo_ground_result_t result, void *data) {
+                static_cast<GroundEventHandler *>(data)->finish(static_cast<GroundResult>(result));
+            },
+            [](void *data) {
+                if constexpr (!std::is_null_pointer_v<HandlerType> && !std::is_pointer_v<Handler> &&
+                              !std::is_lvalue_reference_v<Handler>) {
+                    std::unique_ptr<HandlerType>(static_cast<HandlerType *>(data));
```
BenKaufmann: Consider using `std::default_delete` here, which more directly expresses the intend.

lib/cxx-api/include/clingo/control.hh
rkaminsk: This would make things much simpler. I'll give it a shot.

lib/c-api/src/ground.cc
```diff
@@ -0,0 +1,130 @@
+#include <clingo/util/algorithm.hh>
+
+#include <clingo/control.h>
+#include <clingo/script.h>
+
+#include <clasp/cli/clasp_options.h>
```
BenKaufmann: Is this header actually needed here? Same for the program-opts headers.

lib/c-api/src/lib.hh
```diff
@@ -23,6 +23,48 @@ struct clingo_lib {
...
+    ~CHandler() { cleanup(); }
+
+  private:
+    void cleanup() noexcept {
+        if (handler_ && handler_->free) {
+            handler_->free(data_);
+        }
+    }
+
+    std::optional<Handler> handler_;
```
BenKaufmann, line 64: What's the motivation for using an optional instead of directly using `nullptr` as the "none" value?

lib/control/src/solver.cc
```diff
@@ -1206,12 +1207,110 @@ auto Solver::const_map() -> Input::ConstMap const & {
...
+
+    void cancel() {
+        if (future_.valid()) {
+            stop_.request_stop();
+            future_.wait();
+        }
+    }
+
+    ~Impl() noexcept {
+        try {
```
BenKaufmann, line 1271: If you are not planing to log or otherwise handle the exception, I would suggest dropping the explicit `try-catch-terminate` as it would happen implicitly anyway.

lib/cxx-api/include/clingo/control.hh
```diff
@@ -220,6 +283,110 @@ struct ProfileNodeInternal {
     std::vector<ProfileNode> children;
 };
 
+//! Enumeration of the results of a grounding process.
+enum class GroundResult : clingo_ground_result_t {
```
BenKaufmann: How is this different from the `GroundResult` in `core.hh`?

lib/util/include/clingo/util/sync.hh
```diff
@@ -0,0 +1,33 @@
+#pragma once
```
BenKaufmann, line 1: Maybe consider renaming this file as it is not really about threads.

lib/c-api/src/lib.hh
rkaminsk: The C struct is stored by value not pointer. I thought to copy it to avoid any lifetime issues here.

lib/control/src/solver.cc
rkaminsk: I should probably emit some error message here. This terminate is extremely unlikely anyway because there should be no way cancel can throw.

lib/cxx-api/include/clingo/control.hh
rkaminsk: It is not different at all. It's just that the two involved libraries only have an internal dependency. The headers of the internal one are not even installed. Thus, there is a bit of duplication.

lib/control/src/grounder.cc
rkaminsk: Done.

lib/cxx-api/include/clingo/control.hh
rkaminsk: I used std::default_delete. My idea was to make the round trip via unique pointer complete here because this is also how the life of the pointer started.

lib/c-api/src/ground.cc
rkaminsk: Probably left overs from refactorings. C++ is at fault not me. :grin:

lib/c-api/include/clingo/control.h
```diff
@@ -200,14 +137,39 @@ CLINGO_VISIBILITY_DEFAULT bool clingo_control_parse_string(clingo_control_t *con
...
+//! @param[in] handler handler for grounding events
 //! @param[in] data user data for ground_callback
 //! @return whether the call was successful
 CLINGO_VISIBILITY_DEFAULT bool clingo_control_ground(clingo_control_t *control, clingo_part_t const *parts, size_t size,
-                                                     clingo_ground_callback_t ground_callback, void *data);
+                                                     clingo_ground_event_handler_t const *handler, void *data);
+
+//! Start grounding a program asynchronously.
+//!
+//! The function returns a handle and immediatly starts grounding in the
```
BenKaufmann: Nitpick: _immediatly_ -> immediately

lib/cxx-api/include/clingo/control.hh
```diff
@@ -310,7 +241,7 @@ class Control {
         Detail::handle_error(clingo_control_parse_string(ctl_.get(), program.data(), program.size()));
     }
 
-    //! Ground the control object with the given parts.
+    //! Ground the logic program with the given parameters.
     //!
     //! The given parts determine which program parts are grounded with which
     //! paramaters. If no parts are given, the parts given by the parts
```
BenKaufmann: Nitpick: _paramaters_ -> parameters

lib/c-api/include/clingo/ground.h
```diff
@@ -0,0 +1,177 @@
...
+//!
+//! @param[in] handle the target
+//! @param[out] result the ground result
+//! @return whether the call was successful
+CLINGO_VISIBILITY_DEFAULT bool clingo_ground_handle_get(clingo_ground_handle_t *handle, clingo_ground_result_t *result);
+//! Wait for the specified amount of time to check if the result is ready.
+//!
+//! If the time is set to zero, this function can be used to poll if grounding
+//! is still active. If the time is negative, the function blocks until the
+//! groinding is finished.
```
BenKaufmann: Nitpick: _groinding_ -> grounding

lib/c-api/src/ground.cc
```diff
@@ -0,0 +1,125 @@
...
+        CLINGO_CATCH;
+    }
+
+    clingo_lib_t *lib_ = nullptr;
+    CHandler<clingo_ground_event_handler_t> handler_;
+};
+
+auto cpp_cast(clingo_ground_handle_t *hnd) -> CppClingo::Control::GroundHandle * {
+    if (hnd == nullptr) {
+        throw std::runtime_error("ground handle is null");
```
BenKaufmann: This looks more like a logic error (precondition violation) to me.

lib/c-api/src/ground.cc
```diff
@@ -0,0 +1,125 @@
...
+    CLINGO_CATCH;
+}
+
+extern "C" auto clingo_control_start_ground(clingo_control_t *control, clingo_part_t const *parts, size_t size,
+                                            clingo_ground_event_handler_t const *handler, void *data,
+                                            clingo_ground_handle_t **handle) -> bool {
+    CLINGO_TRY {
+        auto hnd = CHandler{handler, data};
+        auto ctx = hnd ? std::make_unique<Handler>(control->lib, std::move(hnd)) : nullptr;
+        *handle = c_cast(std::make_unique<CppClingo::Control::GroundHandle>(
```
BenKaufmann, line 120: Maybe consider asserting/checking that `handle` is not null.

lib/c-api/src/lib.hh
```diff
@@ -23,6 +23,48 @@ struct clingo_lib {
...
+        return *this;
+    }
+
+    auto operator->() -> Handler * { return handler(); }
+    operator bool() const noexcept { return handler_.has_value(); }
+
+    auto handler() -> Handler * { return handler_ ? &*handler_ : nullptr; }
+    auto data() -> void * { return data_; }
+
+    ~CHandler() { cleanup(); }
```
BenKaufmann: Nitpick: For consistency, maybe consider moving the dtor up below the ctors.

lib/c-api/src/lib.hh
```diff
@@ -23,6 +23,48 @@ struct clingo_lib {
...
+        if (this != &other) {
+            cleanup();
+            handler_ = std::exchange(other.handler_, std::nullopt);
+            data_ = std::exchange(other.data_, nullptr);
+        }
+        return *this;
+    }
+
+    auto operator->() -> Handler * { return handler(); }
+    operator bool() const noexcept { return handler_.has_value(); }
```
BenKaufmann: I would suggest making this operator explicit to avoid unintentional conversions.

lib/control/include/clingo/control/solver.hh
```diff
@@ -588,6 +588,77 @@ class Grounded : public Clasp::Event {
...
+    //!
+    //! If the timeout is greater than zero, the method blocks for at most
+    //! timeout seconds. If the timeout is zero, the method does not block. If
+    //! the timeout is negative, the method blocks until grounding has
+    //! finished.
+    //!
+    //! The function returns true if grounding has finished, false otherwise.
+    //!
+    //! @param timeout The maximum time to wait in seconds.
+    [[nodiscard]] auto wait(double timeout) -> bool;
```
BenKaufmann, line 646: Out of curiosity: what's your heuristic for deciding on how to represent "no timeout"? E.g. you are using  `< 0` here but `std::optional<double>` in `lib/cxx-api/include/clingo/ground.hh`.

lib/control/include/clingo/control/solver.hh
```diff
@@ -588,6 +588,77 @@ class Grounded : public Clasp::Event {
     std::span<Input::ProgramParamVec::value_type const> params;
 };
 
+class Grounder;
+
+//! This callback interface provides grounding events.
+//!
+//! It extends the script callback interface with an callback that is called
```
BenKaufmann: Nitpick: _an callback_ -> a callback

lib/control/include/clingo/control/solver.hh
```diff
@@ -615,8 +686,11 @@ class Solver : public BaseView {
     void add_const(String name, Symbol value);
     //! Get the const map.
     [[nodiscard]] auto const_map() -> Input::ConstMap const &;
+    //! Ground the program asynchronously.
+    [[nodiscard]] auto start_ground(ProgramParamVec params, UGroundEventHandler handler) -> GroundHandle;
     //! Ground the program.
-    void ground(ProgramParamVec const &params, Ground::ScriptCallback *ctx);
+    auto ground(ProgramParamVec const &params, Ground::ScriptCallback *ctx, Util::StopFlag *stop = nullptr)
```
BenKaufmann, line 692: Do you think it makes sense to check for `stop->stop_requested()` on entry to this function or somewhere else early in the grounding process?

lib/control/src/solver.cc
```diff
@@ -1044,7 +1045,7 @@ void Solver::interrupt() noexcept {
         clasp_facade().interrupt(random_signal);
     } catch (std::exception const &e) {
         printf("panic: %s\n", e.what());
```
BenKaufmann: Maybe consider panicking to stderr.

lib/control/src/solver.cc
```diff
@@ -1206,12 +1207,110 @@ auto Solver::const_map() -> Input::ConstMap const & {
...
+            stop_.request_stop();
+            future_.wait();
+        }
+    }
+
+    ~Impl() noexcept {
+        try {
+            cancel();
+        } catch (std::exception const &e) {
+            printf("panic: %s\n", e.what());
```
BenKaufmann: See comment above. You could also consider providing a `panic(msg)/terminate(msg)` helper function to avoid duplication.

lib/util/include/clingo/util/sync.hh
```diff
@@ -0,0 +1,33 @@
...
+
+#include <atomic>
+
+namespace CppClingo::Util {
+
+//! Helper class to signal stopping of grounding.
+//!
+//! Ideally, std::stop_token would be used instead. Unfortunately, clang 18 and
+//! 19 do not support it yet.
+class StopFlag {
```
BenKaufmann, line 11: Should we document that this is a "one-time" flag? I.e. once it is set, you cannot unset it and instead have to use a new flag for subsequent operations.

lib/cxx-api/include/clingo/control.hh
```diff
@@ -334,8 +265,32 @@ class Control {
...
+            nullptr,
+            nullptr,
+        };
+
         Detail::handle_error(
-            clingo_control_ground(ctl_.get(), c_parts.data(), c_parts.size(), ctx ? &ctx_ : nullptr, &ctx));
+            clingo_control_ground(ctl_.get(), c_parts.data(), c_parts.size(), ctx ? &handler : nullptr, &ctx));
     }
 
     //! Ground the control object with the given parts.
```
BenKaufmann, line 281: Nitpick: _paramaters_ -> parameters

lib/cxx-api/include/clingo/control.hh
```diff
@@ -349,6 +304,82 @@ class Control {
         ground(PartSpan{parts}, std::move(ctx));
     }
 
+    //! Ground the logic program with the given parameters.
+    //!
+    //! The given parts determine which program parts are grounded with which
+    //! paramaters. If no parts are given, the parts given by the parts
```
BenKaufmann: Nitpick: _paramaters_ -> parameters

lib/cxx-api/include/clingo/control.hh
```diff
@@ -349,6 +304,82 @@ class Control {
...
+        if (parts) {
+            c_parts.reserve(parts->size());
+            for (auto const &part : *parts) {
+                c_parts.emplace_back(part.name.data(), part.name.size(), c_cast(part.params.data()),
+                                     part.params.size());
+            }
+        } else {
+            c_parts.reserve(1);
+            c_parts.emplace_back("base", 4, nullptr, 0);
+        }
```
BenKaufmann: Consider extracting this into its own function so that it can be called from here and `ground()` above.

lib/cxx-api/include/clingo/core.hh
```diff
@@ -304,6 +304,87 @@ template <class T, class P> class intrusive_handle {
...
+    //! Free the managed data.
+    static auto free(void *data) { std::default_delete<ValueType>()(cast(data)); }
+
+  private:
+    std::unique_ptr<ValueType> owned;
+    PointerType data = nullptr;
+};
+
+//! Specialization for nullptr_t.
+template <> class UserDataManager<std::nullptr_t> {
```
BenKaufmann: You could drop this specialization by adding a `UserDataManager(std::nullptr_t)` overload to the base template (calling delete on a nullptr is well-defined).

lib/cxx-api/include/clingo/control.hh
```diff
@@ -349,6 +304,82 @@ class Control {
...
+    //! Ground the logic program with the given parameters.
+    //!
+    //! The given parts determine which program parts are grounded with which
+    //! paramaters. If no parts are given, the parts given by the parts
+    //! directive are grounded.
+    //!
+    //! A handler implementing the GroundEventHandler interface can be given to
+    //! handle grounding events. The handler can be passed by
+    //! value/reference/pointer. If passed by value, the handler must be
+    //! copyable. If passed by reference/pointer, lifetime must be managed by
```
BenKaufmann: It seems that this documentation is outdated as `Handler` can't be a pointer.

lib/cxx-api/include/clingo/core.hh
```diff
@@ -304,6 +304,87 @@ template <class T, class P> class intrusive_handle {
     P *ptr_ = nullptr;
 };
 
+//! Manager for user data passed to C callbacks.
+//!
+//! The user data manager handles ownership and casting of user data from void
+//! pointers.
+template <typename Value> class UserDataManager {
```
BenKaufmann: I find this construction a little weird as it only make sense in a very specific context (e.g. it would be very bad to call `free()` while the UserDataManager still owns the object) but given that this is in "Detail", it is probably fine.

lib/cxx-api/include/clingo/ground.hh
```diff
@@ -0,0 +1,140 @@
...
+    //!
+    //! @param timeout the optional timeout in seconds
+    //! @return whether a model or result is available
+    [[nodiscard]] auto wait(std::optional<double> timeout) -> bool {
+        return Detail::call<clingo_ground_handle_wait>(hnd_.get(), timeout.value_or(-1));
+    }
+
+  private:
+    struct Del {
+        void operator()(clingo_ground_handle_t *hnd) const { clingo_ground_handle_close(hnd); }
```
BenKaufmann: Nitpick: Consider making this `noexcept` - it is not really needed but this is what e.g. `std::default_delete` does.

lib/c-api/src/ground.cc
rkaminsk: Yeah, would be better and I could start doing it here. Maybe I'll do this at some point in a separate PR where I go over the whole API.

lib/cxx-api/include/clingo/core.hh
rkaminsk: I think, I keep the current version. It seems more straightforward and avoids unnecessary members.

lib/cxx-api/include/clingo/core.hh
rkaminsk: I did not have a better idea. The data has to be released and control handed over to the C API at somepoint. The current construct allows for creating as many user data objects as required. The manager is exception safe until the data is released and handed to the C API at which point the C function is in charge of calling free at the appropriate time. I plan on reusing this construct in other places where the C API takes owner ship.

lib/cxx-api/include/clingo/control.hh
```diff
@@ -587,6 +576,26 @@ class Control {
         return SolveHandle{res};
     }
 
+    [[nodiscard]] auto c_parts_(std::optional<PartSpan> parts = std::nullopt) const
```
BenKaufmann, line 713: You could slightly simplify this function by making the ctor of `Part`  constexpr.
```cpp
if (!parts) {
  static constexpr auto base_part = Part{"base", {}}; // constexpr - no alloc 
  default_parts = this->parts();
  parts = default_parts ? *default_parts : PartSpan{&base_part, 1};
}
c_parts.reserve(parts->size());
for (auto const &part : *parts) {
  c_parts.emplace_back(part.name.data(), part.name.size(), c_cast(part.params.data()), part.params.size());
}
```

lib/cxx-api/include/clingo/control.hh
```diff
@@ -319,23 +250,32 @@ class Control {
...
-            c_parts.reserve(parts->size());
-            for (auto const &part : *parts) {
-                c_parts.emplace_back(part.name.data(), part.name.size(), c_cast(part.params.data()),
-                                     part.params.size());
-            }
-        } else {
-            c_parts.reserve(1);
-            c_parts.emplace_back("base", 4, nullptr, 0);
-        }
+        auto [default_parts, c_parts] = c_parts_(parts);
```
BenKaufmann, line 253: FYI: It took me a minute to understand the role of the unused _default_parts_ element as a (scoped) "lifetime-extender". I don't have a good suggestion but maybe you could come up with a better name.

lib/cxx-api/include/clingo/core.hh
BenKaufmann: I have other ideas but they are probably not objectively better, so I won't further annoy you :smile: 

However, if you plan to reuse `UserDataManager`, I would suggest implementing (or disabling) the move operations so that `data` is also cleared on move.

lib/cxx-api/include/clingo/core.hh
```diff
@@ -304,6 +304,87 @@ template <class T, class P> class intrusive_handle {
...
+        return data;
+    }
+
+    //! Cast the managed data to the pointer type.
+    static auto cast(void *data) -> PointerType { return static_cast<PointerType>(data); }
+    //! Free the managed data.
+    static auto free(void *data) { std::default_delete<ValueType>()(cast(data)); }
+
+  private:
+    std::unique_ptr<ValueType> owned;
```
BenKaufmann: Nitpick: The private members should probably end with an `_`.

lib/cxx-api/include/clingo/control.hh
rkaminsk: The part is using a std::string. So this cannot be constexpr, right?

lib/cxx-api/include/clingo/core.hh
rkaminsk: I can delete move/copy operations. The idea is to not/copy move these around anyways. I even had them written down once. It's just that these 4 lines are so ugly and I have to repeat them 3 times. But that's how it is...

lib/cxx-api/include/clingo/control.hh
rkaminsk: I just add a line to the docs.

lib/cxx-api/include/clingo/core.hh
rkaminsk: If you have a nicer way to write this, I am all ears. I was struggling quite a bit to come up with names and keep keep everything in one place.

lib/cxx-api/include/clingo/control.hh
```diff
@@ -349,6 +289,67 @@ class Control {
...
+                },
+                [](clingo_ground_result_t result, void *data) {
+                    UserData::cast(data)->finish(static_cast<GroundResult>(result));
+                },
+                [](void *data) { UserData::free(data); },
+            };
+            c_handler_ptr = &c_handler;
+        }
+        clingo_ground_handle_t *handle = nullptr;
+        Detail::handle_error(clingo_control_start_ground(ctl_.get(), c_parts.data(), c_parts.size(),
```
BenKaufmann: Please consider using `c_handler_ptr` directly here. I.e.:
```cpp
Detail::handle_error(clingo_control_start_ground(ctl_.get(), c_parts.data(), c_parts.size(), c_handler_ptr,
                                                         user_data.release(), &handle));
```
It has the same effect but is more obvious and avoids confusion (as written the code would be incorrect if one would (naively) assume `user_data.release()` behaves like  `unique_ptr::release`).

lib/cxx-api/include/clingo/control.hh
BenKaufmann: > The part is using a std::string. So this cannot be constexpr, right?

No. Both std::string and std::vector are (limited) constexpr since C++20 and it works with clang>=18, gcc>=13, and msvc.

lib/cxx-api/include/clingo/control.hh
rkaminsk: Yes, in this variant it is possible to directly use c_handler_ptr. It was actually not possible in the variants before but after refactoring this function a hundred times it is possible. With this variant, I could even remove the separate data from the UserDataManager. I think, I'll do exactly this.

lib/cxx-api/include/clingo/core.hh
BenKaufmann: > If you have a nicer way to write this, I am all ears. I was struggling quite a bit to come up with names and keep keep everything in one place.

I would probably put everything in one template class and give it a little more "unique_ptr" semantics . But again, I don't claim that this is better:
```cpp
//! Manager for user data passed to C callbacks.
//!
//! The user data manager handles ownership and casting of user data from void pointers.
template <typename Value> class UserDataManager {
  private:
    static_assert(!std::is_reference_v<Value> && !std::is_pointer_v<Value> && !std::is_array_v<Value>, "sanity check");
    // NOTE: The following could also be factored out into a separate ValueHolder traits template (via specialization).
    static constexpr auto no_del = [](auto *) {};
    static constexpr auto is_ref = std::is_reference_v<std::unwrap_reference_t<Value>>;
    using HolderType =
        std::conditional_t<!is_ref, std::unique_ptr<Value>,
                           std::unique_ptr<std::remove_reference_t<std::unwrap_reference_t<Value>>, decltype(no_del)>>;

  public:
    using ValueType = HolderType::element_type;
    using PointerType = HolderType::pointer;
    explicit UserDataManager(std::nullptr_t) {}
    //! Construct the managed user data from the given value.
    template <class T>
    explicit UserDataManager(T &&value)
        requires(!is_ref)
        : data_(std::make_unique<ValueType>(std::forward<T>(value))) {}
    template <class T>
    UserDataManager(std::reference_wrapper<T> x)
        requires(is_ref)
        : data_(&x.get()) {}

    //! Check whether the managed data is not null.
    explicit operator bool() const { return data_ != nullptr; }
    //! Get the managed data and release ownership.
    auto release() -> PointerType { return data_.release(); }
    // NOTE: Could also add a get()  functio

    //! Cast the managed data to the pointer type.
    static auto cast(void *data) -> PointerType { return static_cast<PointerType>(data); }
    //! Free the managed data.
    static auto free(void *data) { typename HolderType::deleter_type{}(cast(data)); }

  private:
    HolderType data_;
};
```

lib/cxx-api/include/clingo/control.hh
rkaminsk: Just so that I do not get anything wrong. We cannot write
```
constexpr std::string x = "value";
```
because this would require a constexpr allocator. Even thought I could make the Part constructor constexpr, this would only have an effect if I would pass in a constexpr string (like the empty one). So your code above cannot work.

lib/cxx-api/include/clingo/core.hh
rkaminsk: So essentially the same interface with a slightly more compact implementation.

Btw. will zero size optimization apply in

    std::unique_ptr<std::remove_reference_t<std::unwrap_reference_t<Value>>, decltype(no_del)>

to avoid blowing up the unique pointer? (I would guess yes.)

lib/cxx-api/include/clingo/control.hh
BenKaufmann: > Just so that I do not get anything wrong. We cannot write
> 
> ```
> constexpr std::string x = "value";
> ```
> 
> because this would require a constexpr allocator. 

You are wrong. This only requires an allocation for strings whose length exceeds the SSO and for all relevant implementations SSO-length is >= 5 (required for "base").
C.f: https://godbolt.org/z/j57x5Mxsa

lib/cxx-api/include/clingo/core.hh
BenKaufmann: > So essentially the same interface with a slightly more compact implementation.

Yes. I also played around with alternative approaches but given that I don't know the exact requirements (e.g. regarding when to actually do the allocation/copying), I simply stayed with the interface.

> Btw. will zero size optimization apply in

Yes. `static_assert(sizeof(void *) == sizeof(HolderType));` is required to be true here.

lib/cxx-api/include/clingo/control.hh
rkaminsk: Shouldn't we avoid relying on an unspecified SSO-length?

rkaminsk: As a compromise, I could write

static auto default_part = Part{"base", {}};

There will also be no allocation.

BenKaufmann: > Shouldn't we avoid relying on an unspecified SSO-length?

To me, this kind of "unspecified" is harmless. It either compiles and works as expected or it fails to compile with a clear error message.

> As a compromise, I could write

Please don't feel pressured to compromise. This was just meant as an FYI.

lib/cxx-api/include/clingo/core.hh
```diff
@@ -304,6 +304,101 @@ template <class T, class P> class intrusive_handle {
...
+
+template <typename T> struct UserDataTraitsBase<std::unique_ptr<T>> {
+    using ValueType = T;
+    using PtrType = ValueType *;
+    using HolderType = std::unique_ptr<ValueType>;
+    static constexpr auto type = UserDataType::ptr;
+};
+
+template <> struct UserDataTraitsBase<std::nullptr_t> {
+    using ValueType = nullptr_t;
```
BenKaufmann: For consistency, I would suggest either adding `std::` here or dropping the `std::` in the line above.

lib/cxx-api/include/clingo/core.hh
```diff
@@ -304,6 +304,101 @@ template <class T, class P> class intrusive_handle {
     P *ptr_ = nullptr;
 };
 
+enum class UserDataType {
```
BenKaufmann: Style: You could slightly reduce duplication and get rid of the additional type enum by combining the approach from `std::iterator` and transparent comparators.
```cpp
// Common type with required typedefs.
template <typename HT, typename PT = HT> struct UserDataTraitsHolder {
    using HolderType = HT;
    using PtrType = PT;
};
// Base template - for value types
template <typename T> struct UserDataTraitsBase : UserDataTraitsHolder<std::unique_ptr<T>, T *> {
    using ValueType = T; // indicate that this is a value type (similar to is_transparent)
};
// Specializations for "non"-value types (-> no ValueType member)
template <typename T> struct UserDataTraitsBase<std::reference_wrapper<T>> : UserDataTraitsHolder<T *> {};
template <typename T> struct UserDataTraitsBase<std::unique_ptr<T>> : UserDataTraitsHolder<std::unique_ptr<T>, T *> {};
template <> struct UserDataTraitsBase<std::nullptr_t> : UserDataTraitsHolder<std::nullptr_t> {};
```

Functions `create()`, `release()`, and `free()` then use the provided interface/typedefs to select the appropriate action. E.g. 
```cpp
...
static constexpr bool has_data = not std::is_null_pointer_v<PtrType>;
static constexpr bool is_value = requires { typename BaseType::ValueType; };
template <typename U> static auto create(U &&data) -> HolderType {
    if constexpr (is_value) { // value type?
        return std::make_unique<typename BaseType::ValueType>(std::forward<U>(data));
    } else if constexpr (requires {
                             { &data.get() } -> std::same_as<HolderType>;
                         }) { // reference-wrapper? Could also using something like `typename BaseType::is_ref`
        return &data.get();
    } else {
        return data;
    }
}
```

rkaminsk: What do you think about the following?

```c++
template <typename HT, typename PT> struct UserDataTraitsHolder {
    using HolderType = HT;
    using PtrType = PT;
};
template <typename T> struct UserDataTraitsBase : UserDataTraitsHolder<std::unique_ptr<T>, T *> {
    using has_release = void;
    using is_ptr = void;
};
template <typename T> struct UserDataTraitsBase<std::reference_wrapper<T>> : UserDataTraitsHolder<T *, T *> {
    using has_release = void;
};
template <typename T> struct UserDataTraitsBase<std::unique_ptr<T>> : UserDataTraitsHolder<std::unique_ptr<T>, T *> {
    using is_ptr = void;
};
template <> struct UserDataTraitsBase<std::nullptr_t> : UserDataTraitsHolder<std::nullptr_t, std::nullptr_t> {};

//! Manager for user data passed to C callbacks.
//!
//! Manages ownership semantics and type-safe casting for user data stored as
//! void pointers. By default, user data is forwarded into a holder object.
//! Reference wrappers are stored as raw pointers, avoiding unnecessary copying
//! or moving of the underlying data. Types representing null pointers are
//! explicitly marked to indicate the absence of user data.
template <typename T> struct UserDataTraits {
    using Base = UserDataTraitsBase<std::remove_cvref_t<T>>;

    //! Create the holder from the given data.
    template <typename U> static auto create(U &&data) -> Base::HolderType {
        if constexpr (requires { Base::is_value; }) {
            return std::make_unique<typename Base::ValueType>(std::forward<U>(data));
        } else if constexpr (requires { Base::is_ref; }) {
            return &data.get();
        } else {
            return data;
        }
    }
    //! Release the pointer to the value from the holder.
    static auto release(Base::HolderType &holder) -> Base::PtrType {
        if constexpr (requires { Base::has_release; }) {
            return holder.release();
        } else {
            return holder;
        }
    }
    //! Cast the void pointer to the pointer type.
    static auto cast(void *ptr) -> Base::PtrType { return static_cast<Base::PtrType>(ptr); }
    //! Free the value pointed to by the void pointer.
    static void free(void *ptr) {
        if constexpr (requires { Base::has_release; }) {
            std::default_delete<typename Base::ValueType>{}(cast(ptr));
        }
    }
};
```

rkaminsk: I wanted thinks to be a little more explicit.

rkaminsk: I messed up some usings but those are easy to fix.

lib/control/src/solver.cc
rkaminsk: It is kind of difficult to provide something usable with printf. Since the duplication is minimal, I just stick to the current version.

lib/util/include/clingo/util/sync.hh
rkaminsk: I'd rather not and just add another function to reset when required. For now I did not because I simply did not need one.

lib/control/include/clingo/control/solver.hh
rkaminsk: The cxx-api is the public API so I wanted it to be a bit nicer here. For the internal API I don't care so much.

lib/control/include/clingo/control/solver.hh
rkaminsk: It would probably be very hard to get a stop request in right at the start. Maybe if you run with valgrind or on a single core. There are more places where one could think about interrupting. For now, I'd rather wait for requests from users where the grounding does not stop fast enough.

lib/cxx-api/include/clingo/core.hh
rkaminsk: I am using an adjusted version of the above for now. I also use the user data when registering propagators and scripts which also need lifetime management.

lib/control/src/solver.cc
BenKaufmann: > It is kind of difficult to provide something usable with printf.

I really meant something simple/stupid similar to `perror()`, e.g.
```cpp
void panic(const char* msg) {
  fprintf(stderr, "panic at the disco: %s - going to terminate\n", msg ? msg : "");
  fflush(stderr);
  std::terminate();
}
```
> Since the duplication is minimal, I just stick to the current version.

That's also fine :smile:

rkaminsk:
@BenKaufmann, I hope this is done now. Don't feel obliged to look at this again because the PR got quite large.

PR #595, opened 2026-01-26
fix: do not gobble exceptions during rewriting (C API)

rkaminsk:


lib/input/include/clingo/input/rewrite/rewrite_context.hh
```diff
@@ -173,6 +173,13 @@ class RewriteContext {
     //! more than one error indicator to the user.
     void set_error() { has_error_ = true; }
 
+    //! Clear the error flag and return its previous value.
+    [[nodiscard]] auto clear_error() -> bool {
+        auto ret = has_error_;
```
BenKaufmann: You could consider using `std::exchange()` here

rkaminsk: Of course.

PR #599, opened 2026-02-06
Extend ground output options wip

FrancoisLaferriere:
## Pull request overview

This PR extends clingo's output / conversion capabilities by introducing a configurable backend output format (`--convert=aspif/smodels/reify/text`) and wiring solver output through a stream adapter so the output can also be accessed via the API.

**Changes:**
- Add `BackendType`/`ReifyFlags` to solver options and route solving-mode output through Potassco backends selected via `--convert`.
- Introduce `Util::OutputStream` to bridge `OutputBuffer` with `std::ostream`-based APIs.
- Update docs/stubs/tests to use `--convert=text` and add Python tests covering `--convert` outputs.


lib/control/src/solver.cc
```diff
@@ -252,6 +255,67 @@ class AbstractProgramBackendImpl : public ProgramBackend, public TheoryBackend {
     std::vector<Potassco::WeightLit> wlits_;
 };
 
+template <class T>
+concept PotasscoAbstractProgram = std::derived_from<T, Potassco::AbstractProgram>;
+
+template <PotasscoAbstractProgram T> class AbstractProgramAdapter : public AbstractProgramBackendImpl {
```
rkaminsk: - This is not really an adapter.  Maybe call it PotasscoBackend or similar.
- Maybe we want to avoid the template here. A double virtual call should not be that expensive. Ownership can be managed either via a unique_ptr or be managed externally.

lib/control/src/solver.cc
```diff
@@ -252,6 +255,67 @@ class AbstractProgramBackendImpl : public ProgramBackend, public TheoryBackend {
...
+  public:
+    template <class... U>
+    AbstractProgramAdapter(TermBaseMap &terms, Util::OutputBuffer &out, U &&...args)
```
rkaminsk: We should probably not use the output buffer for now and just use std::cout as output. If we really wanted to, we could implement a stream that forwards to the output buffer. Implementing streams is usually somewhat unpleasant.

lib/control/include/clingo/control/solver.hh
```diff
@@ -64,6 +64,7 @@ enum class AppMode : uint8_t {
     parse,   //!< Stop processing after parsing.
     rewrite, //!< Stop processing after rewriting.
     ground,  //!< Stop processing after grounding.
+    aspif,   //!< TODO
```
rkaminsk: Maybe solve should rather be divided into different solving modes. I'll think about it.

rkaminsk (changes requested):

lib/control/src/solver.cc
BenKaufmann: > * Maybe we want to avoid the template here.

To me, the template approach does not really make sense. `Potassco::AbstractProgram` is an interface (ABC) - this typically implies reference and not value semantics. Additionally, creating and storing a value in `AbstractProgramBackendImpl` creates an unnecessary dependency to the ctor (e.g. we assume that the first argument is always an std::ostream). There is also a (possible) lifetime bug in the code because `prg_` is passed to `AbstractProgramBackendImpl` before it is constructed. 

Maybe I'm missing something but why does `AbstractProgramAdapter` have to  own an `AbstractProgram`? Why can't it just pass on a reference to an `Potassco::AbstractProgram` to its base (AbstractProgramBackendImpl) and then have a (protected) getter function there?

lib/control/src/solver.cc
```diff
@@ -252,6 +258,43 @@ class AbstractProgramBackendImpl : public ProgramBackend, public TheoryBackend {
     std::vector<Potassco::WeightLit> wlits_;
 };
 
+class PotasscoBackend : public AbstractProgramBackendImpl {
+  public:
+    template <class... U>
```
rkaminsk: This has to be removed.

lib/control/src/solver.cc
rkaminsk: I think @FrancoisLaferriere wanted to avoid storing the aspif writer in the solver. There is however no problem doing that. The last change seems fine to me.

rkaminsk: > There is also a (possible) lifetime bug in the code because prg_ is passed to AbstractProgramBackendImpl before it is constructed.

I am guilty of doing such kind of small hacks as well. There should be no lifetime issues as the lifetime of all involved constructs spans the lifetime of the solver object. Since we now have a unique pointer for the program, we can also refactor the ProgramBackendImpl class to use it as well, getting rid of the adapter member.

lib/control/include/clingo/control/solver.hh
rkaminsk: I think it is easier not to introduce another mode. Rather introduce another option to configure the "solving" backend. Then it should suffice to adjust the solve method to not call solve on the facade, if for example the aspif backend has been selected. The advantage is that all the plumbing processes the usual way, only the solver won't receive any rules. From a high level point option point of view we can still introduce an aspif/smodels/reify mode but set the actual mode to solve and the backend to clasp/aspif/smodels/reify respectively.

lib/control/src/solver.cc
rkaminsk: Fine with me now.

lib/control/include/clingo/control/solver.hh
FrancoisLaferriere: I added aspif/smodels/reify as values to the `--mode` option, but did not introduce new `AppMode` variants. Instead, they are internally mapped to  `AppMode::solve` combined with the respective `ModeFormat`.

rkaminsk:
After a quick glance, this is starting to look good. I'll take a closer look later and come back to you.

lib/control/src/solver.cc
```diff
@@ -252,10 +258,46 @@ class AbstractProgramBackendImpl : public ProgramBackend, public TheoryBackend {
...
+    auto do_next_lit() -> prg_lit_t override {
+        if (auto lit = new_atom(); std::cmp_less_equal(lit, prg_lit_max)) {
+            return static_cast<prg_lit_t>(lit);
+        }
+        throw std::range_error("number of literals exhausted");
+    }
+
+    auto do_fact_lit() -> std::optional<prg_lit_t> override {
+        return std::nullopt;
+    } // NOTE: only called in the aspif parser
```
BenKaufmann: Maybe consider dropping this comment or elaborate why this information is important.

lib/control/src/solver.cc
```diff
@@ -252,10 +258,46 @@ class AbstractProgramBackendImpl : public ProgramBackend, public TheoryBackend {
...
+
+    auto do_term_id(Symbol sym) -> prg_id_t override {
+        auto nId = terms_->add(sym, [&, this]() {
```
BenKaufmann: Consider dropping the explicit `this` capture as it is already implied by `&`. 

Nitpick style: Consider dropping the additional `nId` variable - just return the result of the `add()` call directly.

lib/control/src/solver.cc
```diff
@@ -252,10 +258,46 @@ class AbstractProgramBackendImpl : public ProgramBackend, public TheoryBackend {
...
+        });
+        return nId;
+    }
+
+    void do_term_id(Symbol sym, prg_id_t id) override { terms_->add(sym, id); }
+
+    auto new_atom() -> Potassco::Atom_t { return nextAtom_++; }
+    auto new_show_term() -> prg_id_t { return nextShowTerm_++; }
+
+    Potassco::Atom_t nextAtom_{1};
```
BenKaufmann: Nitpick sytle: It looks to me like `clingo` uses `snake_case` for variables. Hence, consider renaming to `next_atom_` and `next_show_term_`

lib/control/src/solver.cc
```diff
@@ -1069,23 +1110,52 @@ auto SymbolTable::output(CppClingo::Symbol const &sym) -> State & {
...
+        case ModeFormat::reify: {
+            Potassco::Reifier::Options reify_opts{};
+            reify_opts.reifyStep = Potassco::test(reify, ReifyFlag::reify_step);
+            reify_opts.calculateSccs = Potassco::test(reify, ReifyFlag::reify_scc);
+            program_ = std::make_unique<Potassco::Reifier>(std::cout, reify_opts);
+            break;
         }
         default: {
-            return Output::make_text_output(buf_);
+            POTASSCO_ASSERT_NOT_REACHED("invalid output format for solving mode");
```
BenKaufmann: Nitpick style: Assuming the goal is to handle all cases (all mode formats), I would drop the `default` case and instead `POTASSCO_ASSERT(program_, "invalid output format for solving mode");` after the switch. This way, the compiler can warn about missing cases thereby turning a  run-time into a compile-time issue.

lib/c-api/src/opts.hh
```diff
@@ -237,7 +237,34 @@ class ClingoOptions {
         }
     }
 
+    template <class ModeT> auto parse_format(std::string_view str, ModeT &mode) -> bool {
```
BenKaufmann: I'm not a huge fan of this template function - the name/signature does not match its implementation.

I would at least a) try to find a better/more obvious name (indicating that the function actually changes internal state) and b) split it into a (mode-independent) parse/apply function and a template. I.e.
```cpp
// TBD: Find good name
auto parse_format(std::string_view str) -> bool {
        namespace Parse = Potassco::Parse;
        using namespace std::literals;
       ...
}
template <class ModeT> auto parse_format(std::string_view str, ModeT &mode) -> bool {
  if (parse_format(str)) {
    mode = ModeT::solve;
    return true;
  }
  return false;
}
```

lib/c-api/src/opts.hh
```diff
@@ -237,7 +237,34 @@ class ClingoOptions {
...
+                    break;
+                }
+            }
+        }
+        return str.empty();
+    }
+
     auto mode() -> Control::AppMode & { return solver_opts_.mode; }
+    auto format() -> Control::ModeFormat & { return solver_opts_.format; }
+    auto reify() -> Control::ReifyFlag & { return solver_opts_.reify; }
```
rkaminsk: Better use `reify_flags`?

lib/c-api/src/opts.hh
```diff
@@ -237,7 +237,34 @@ class ClingoOptions {
...
+                } else {
+                    break;
+                }
+            }
+        }
+        return str.empty();
+    }
+
     auto mode() -> Control::AppMode & { return solver_opts_.mode; }
+    auto format() -> Control::ModeFormat & { return solver_opts_.format; }
```
rkaminsk: What about using output format as name.

lib/c-api/src/opts.hh
rkaminsk: We could use parse_backend_type as name.

lib/control/include/clingo/control/solver.hh
```diff
@@ -67,10 +67,30 @@ enum class AppMode : uint8_t {
     solve    //!< Stop processing after solving.
 };
 
+//! Enumeration of available output formats when in solving mode.
+enum class ModeFormat : uint8_t {
```
rkaminsk: Maybe use BackendType?

lib/control/include/clingo/control/solver.hh
```diff
@@ -67,10 +67,30 @@ enum class AppMode : uint8_t {
...
+    solve_default, //!< Standard solving mode: use the normal backend (Clasp)
```
rkaminsk: We could just call this clasp.

lib/control/include/clingo/control/solver.hh
```diff
@@ -67,10 +67,30 @@ enum class AppMode : uint8_t {
...
+    reify_step = 2U //!< Add step numbers
+};
+POTASSCO_ENABLE_BIT_OPS(ReifyFlag);
+
 //! Options for the solver.
 struct SolverOptions {
     //! Operation mode of the solver.
     AppMode mode = AppMode::solve;
+    //! Output format to use when in solving mode.
+    ModeFormat format = ModeFormat::solve_default;
```
rkaminsk: backend_type

lib/control/src/solver.cc
rkaminsk: We should also test this for the aspif parser, then.

lib/control/src/solver.cc
rkaminsk: Yes, and let's stick with snake_case in the clingo space!

lib/control/include/clingo/control/solver.hh
```diff
@@ -67,10 +67,30 @@ enum class AppMode : uint8_t {
...
+enum class ModeFormat : uint8_t {
+    solve_default, //!< Standard solving mode: use the normal backend (Clasp)
+    aspif,         //!< Print program in ASP intermediate format
+    smodels,       //!< Print program in smodels format
+    reify          //!< Print program as reified facts
+};
+POTASSCO_SET_ENUM_ENTRIES(ModeFormat, {aspif, "aspif"sv}, {smodels, "smodels"sv}, {reify, "reify"sv});
+
+//! Options for reification.
+enum class ReifyFlag : uint8_t {
```
rkaminsk: This should be called `ReifyFlags`.

lib/control/include/clingo/control/solver.hh
```diff
@@ -67,10 +67,30 @@ enum class AppMode : uint8_t {
...
+POTASSCO_ENABLE_BIT_OPS(ReifyFlag);
+
 //! Options for the solver.
 struct SolverOptions {
     //! Operation mode of the solver.
     AppMode mode = AppMode::solve;
+    //! Output format to use when in solving mode.
+    ModeFormat format = ModeFormat::solve_default;
+    //! Reification flags.
+    ReifyFlag reify = {};
```
rkaminsk: `reify_flags`

lib/control/src/solver.cc
rkaminsk: Agreed, we can just drop the default here.

rkaminsk (changes requested):
Let's try to find good names next. Afterward, we will still need to test this.

lib/c-api/src/opts.hh
BenKaufmann: > We could use parse_backend_type as name.

I think the "parse" prefix just does not fit here. In other places, `parse_*` means "transform string to (output) parameter of type T", but here "parse" **both** _applies state_ **and** transforms to output.

From a pure implementation perspective, I would do the following:
1. Provide enum entries for enum class AppMode in `solver.hh`, e.g. via:
```cpp
POTASSCO_REFLECT_ENUM_ENTRIES(AppMode, 0, 4); // or explicitly via POTASSCO_SET_ENUM_ENTRIES
```
2. Implement `auto init_app_mode(std::string_view str) -> bool` in `ClingoOptions`. 
```cpp
auto init_app_mode(std::string_view str) -> bool {
    namespace Parse = Potassco::Parse;
    using namespace std::literals;
    // Parse to temporaries: we don't want partial state changes
    Control::AppMode mode;
    Control::ReifyFlag reify;
    Control::ModeFormat format;
    if (Parse::ok(Potassco::extract(str, format))) {
        if (format == Control::ModeFormat::reify) {
            while (Parse::matchOpt(str, ',')) {
                ...
        }
    } else if (not Parse::ok(Potassco::stringTo(str, mode))) {
        return false;
    }
    if (str.empty()) { // Success: commit result
        solver_opts_.mode = Control::AppMode::solve;
        solver_opts_.format = format;
        solver_opts_.reify = reify;
       return true;
    }
    return false;
}
```
3. In `clingo_control_new`,  bind `--mode` option parser to `init_app_mode`.
```cpp
auto parse_mode = [&opts](std::string_view str) { return opts.init_app_mode(str); };
```
4. In `main.cc:initOptions`, compose `--mode` option parser:
```cpp
auto parse_mode = [this](std::string_view str) {
    if (opts_.init_app_mode(str)) { // "Common" app mode?
        mode_ = static_cast<Mode>(opts_.mode());
        return true;
    }
    if (str == "clasp") { // Special app mode?
        mode_ = Mode::clasp;
        return true;
    }
    return false;
};
```

Anyhow, to me (the naive user), the semantics of the mode option (and its tight coupling to the internal states) is unclear. For example (ignoring implementation), what's the difference between `parse`, `rewrite`, and `ground`, and why is `aspif` etc.  not some kind of rewrite /grounding? 

From a user's perspective, maybe it would be more clear if we'd add a new `convert`/`transform` mode with the output format as sub-arguments.

rkaminsk: We should probably leave the implementation aside for a moment and discuss what the options from a user's perspective should look like:

Currently, we have `--mode={parse,rewrite,ground,solve,clasp}`:
- `parse`: Parses the program and prints it as parsed. Limited use maybe pretty-printing.
- `rewrite`: Rewrites the parsed program. This applies simplifications and brings rules into a shape ready for grounding. Mostly for debugging purposes.
- `ground`: This is essentially clingo's previous `--text` option.
- `solve`: This is the usual clingo ground and solve mode.
- `clasp`: This turns clingo into clasp.

Next we want to add `aspif`/`smodels`/`reify`  output. Technically, these proceed a bit like the solve mode (solve terminates an aspif/smodels slice with a zero). They don't really fit nicely in the above list. 

A separate option `--backend={clasp,aspif,smodels,reify}` or maybe `--output` that requires `solve` mode would also be an option.

What are your thoughts?

BenKaufmann: Here is my naive view: we have a mode `clasp` that turns `clingo` into `clasp` and we have a mode `ground`, which basically turns `clingo` into a grounder. The mode `solve` is the actual `clingo` and shouldn't even be needed in the CLI, so I would actually drop it.
Now, `aspif`/`smodels`/`reify` might internally depend on the "solve" mode, but to me, this is a non-obvious implementation detail since no "solving" is actually happening. 

Conceptually, I would think that `aspif`/`smodels`/`reify` belong to the grounder and hence, I would either have `--mode=ground[,<output-fmt>[,<options>]]` or a separate option `--ground-output=aspif,smodels,reify`. In the latter case, I would suggest generating an error if `--ground-output` is used together with an explicit `mode` != `ground`.

rkaminsk: We can also do it like this. I would still keep the solve mode. Just to have it explicitly.

```
--mode={parse|rewrite|ground|solve|clasp}
--ground-format={text|aspif|smodels|reify}
--reify-options=<list {sccs|steps}>
```

Internally, we would have to make sure that aspify, smodels, reify jump through the solving hoops.

As a convenience feature one could automatically switch to grounding mode if `--ground-format` is given without specifying `--mode`.

FrancoisLaferriere: It seems too technical to present `aspif` / `smodels` / `reify` as backend alternatives to `clasp` for standard users.

Conceptually, they are either:

* Alternative output formats for printing the grounded program, or
* An extra step that converts the text-based grounded program.

I see three options:

**1.** Extend `ground`: `--mode=ground,<text|aspif|smodels|reify>`

**2.** A new mode: `--mode=convert,<aspif|smodels|reify>`

**3.** A new option: `--output=<text|aspif|smodels|reify>`  and require `mode=ground` as option, and `mode=solve` internally.

rkaminsk: `convert` would fit into the existing mode option

But I actually prefer a separate option: `--output`/`--convert`/`--ground-format` whatever. I would also let this option switch the mode behind the scenes as required. To keep things simple we can even make the mode and the new option exclusive.

What about

```
--mode={parse|rewrite|solve|clasp}
--output={text|aspif|smodels|reify}
--reify-options=<list {sccs,steps}>
```

?

Option `--output` can set the internal mode to ground/solve as required. If both are given, we can raise an exception. Then mode becomes an advanced option for debugging and using clasp. Option `--output` can be used to inspect the text output and use clingo with other solvers.

I won't insist on this solution if you think there is a better one. :smile:

BenKaufmann: I don't have a strong preference. However, if we go with the separate option, I would prefer a different name for that option. Having both `--output` and (the poorly named clasp option) `--outf` might be confusing. Of course, we could also rename the latter.

rkaminsk: If everyone is okay with having separate options we can at least go forward with the implementation. Changing names of options later should not be a big deal.

FrancoisLaferriere: Let’s settle on a separate option then. What about the reifier options? We could also have a separate option, as Roland suggested, or have them comma-separated, the same way as for `--pre` in clasp. I tend more toward the second option for consistency with clasp.

rkaminsk: I have no strong preference here.

FrancoisLaferriere:
I used `--convert` for now, and used comma-separated option for the reifier.

Todo:
- Remove `ground` from AppMode enum entries. I temporarily kept it to not break the tests, which need update.
- I still need to see what's up with `do_fact_lit()`

rkaminsk:
> I still need to see what's up with `do_fact_lit()`

The function exists for backward compatibility with aspif 1.0. It is rather easy to implement:

- introduce a fact_ member that is initially zero
- if fact is zero and a rule is output that is a fact -> set fact_ to the head of the rule
- if do_fact_lit is called
  - generate a rule that is a fact introducing a new atom if fact_ is zero
  - return fact_

lib/c-api/src/opts.hh
```diff
@@ -171,6 +171,42 @@ class ClingoOptions {
             }
             return true;
         };
+        auto parse_convert = [this](std::string_view str) {
+            namespace Parse = Potassco::Parse;
+            using namespace std::literals;
+
+            if (auto key = "text"sv; Parse::eqIgnoreCase(str, key, key.size())) {
```
BenKaufmann: Consider simplifying to:
```cpp
if (Parse::eqIgnoreCase(str, "text"sv)) {
  solver_opts_.mode = Control::AppMode::ground;
  return true;
}
```
This way, you only change the mode on success while (incorrect9 input like e.g. `textFoo` is handled in the subsequent attempt to extract a backend type.

lib/c-api/src/control.cc
```diff
@@ -75,6 +75,7 @@ extern "C" auto clingo_control_new(clingo_lib_t *lib, clingo_string_t const *arg
         DefaultParseContext pc{ctx};
         parseCommandArray(pc, {cargs.data(), size}, pos_parser);
         ctx.assignDefaults(pc.parsed());
+        opts.validate_options(pc.parsed());
```
BenKaufmann, line 77: Nitpick: I would suggest validating the parsed options before applying defaults.

lib/c-api/src/main.cc
```diff
@@ -174,31 +174,48 @@ class ClingoApp : public Clasp::Cli::ClaspAppBase {
...
-                                              {"parse", Mode::parse},
-                                              {"rewrite", Mode::rewrite},
-                                              {"ground", Mode::ground},
-                                              {"solve", Mode::solve},
-                                              {"clasp", Mode::clasp},
-                                          })),
-             "Run in {parse|rewrite|ground|solve|clasp} mode");
+            ("mode", parse(parse_mode),
+             "Run in the specified mode\n"
+             "      %A: <mode {parse|rewrite|solve|clasp}>\n"
```
BenKaufmann: I think it would be more helpful if the documentation explained what kind of output is produced/action is performed. E.g. "Print parsed program and exit" (similar to clasp's `--pre` option).

lib/c-api/src/opts.hh
```diff
@@ -219,7 +255,16 @@ class ClingoOptions {
       [no-]atom-undefined     : a :- b.
       [no-]file-included      : #include "a.lp". #include "a.lp".
       [no-]operation-undefined: p(1/0).
-      [no-]global-variable    : :- #count { X } = 1, X = 1.)");
+      [no-]global-variable    : :- #count { X } = 1, X = 1.)")                                     //
+            ("convert", parse(parse_convert),
+             "Convert to specified format\n"
```
BenKaufmann: Maybe consider "Convert program to..." or "Convert input to..."

lib/c-api/src/opts.hh
```diff
@@ -237,7 +282,28 @@ class ClingoOptions {
...
+        if (str.empty()) {
+            solver_opts_.mode = mode;
+            return true;
+        }
+        return false;
+    }
+
+    void validate_options(const Potassco::ProgramOptions::ParsedOptions &parsed) {
+        if (parsed.contains("mode") && parsed.contains("convert")) {
+            throw std::invalid_argument("option '--mode' and '--convert' cannot be used together");
```
BenKaufmann: Nitpick: consider throwing a `Potassco::ProgramOptions::Error` instead - this way, you'd get the additional "Try '--help' for usage information" message in the output.

lib/c-api/src/opts.hh
```diff
@@ -237,7 +283,28 @@ class ClingoOptions {
         }
     }
 
+    auto init_app_mode(std::string_view str) -> bool {
+        namespace Parse = Potassco::Parse;
+
+        Control::AppMode mode{};
+        if (not Parse::ok(Potassco::extract(str, mode))) {
```
BenKaufmann: FYI: I think `clingo` does not use `not` but `!`

lib/control/src/solver.cc
```diff
@@ -252,10 +258,56 @@ class AbstractProgramBackendImpl : public ProgramBackend, public TheoryBackend {
...
+
+    auto do_fact_lit() -> std::optional<prg_lit_t> override {
+        if (fact_ != 0) {
+            return fact_;
+        }
+        return std::nullopt;
+    }
+
+    void do_rule(PrgLitSpan head, PrgLitSpan body, bool choice) override {
+        if (fact_ == 0 && not choice && head.size() == 1 && body.empty()) {
```
BenKaufmann: See above regarding `not`.

rkaminsk:
I made a pass over the PR. Some notes:

- merged with the latest wip-20
- fine-tuned options
  - what do you think about handling the options like this?
- rerouted all grounder output through the output buffer
  - we still need some tests for the conversion options 
  - the change should allow to easily write tests that intercept the smodels and aspif output
  - we only need some small sanity checks whether everything is wired correctly
- the aspif/smodels backends can actually generate ids
  - so the do_fact_lit rule can **introduce** a fact if necessary
- plus some further refinements

rkaminsk:
To give an examples how to access the output via the API:
```python

lib = Library()
ctl = Control(lib, ["--convert=aspif"])
ctl.parse_string("1 {a; b} 1.")
ctl.ground()
print(ctl.solve())
print("<" * 80)
print(ctl.buffer, end="")
print(">" * 80)
```
The program prints:
```
SAT
<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
asp 2 0 0 incremental
1 0 1 1 0 0
1 1 1 3 0 2 1 2
1 1 1 4 0 2 1 2
1 0 0 0 2 -7 1
4 0 3 1 a
4 0 4 1 b
1 0 1 8 1 1 2 5 1 6 1
1 0 1 9 1 2 2 5 1 6 1
1 0 1 7 0 2 8 -9
1 0 1 2 0 0
1 0 1 5 0 2 2 3
1 0 1 6 0 2 2 4
0
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
```

rkaminsk:
There also seems to be a bug with smodels conversion:
```
❯ echo '{x; y} = 1.' | clingo --convert=smodels
*** ERROR: (clingo): SmodelsOutput &Potassco::SmodelsOutput::startRule(SmodelsType):339: Precondition 'sec_ == 0 || rt == SmodelsType::end || rt >= SmodelsType::clasp_increment' failed.
*** Info : (clingo): message: adding rules after symbols not supported
```
It seems like this backend assumes that rules are output before symbols. Clingo has not been written with such a requirement in mind. I am not sure what the best fix might be. One idea could be to record output statements and output them a bit later.

BenKaufmann:
> There also seems to be a bug with smodels conversion:
> 
> ```
> 
> ❯ echo '{x; y} = 1.' | clingo --convert=smodels
> 
> *** ERROR: (clingo): SmodelsOutput &Potassco::SmodelsOutput::startRule(SmodelsType):339: Precondition 'sec_ == 0 || rt == SmodelsType::end || rt >= SmodelsType::clasp_increment' failed.
> 
> *** Info : (clingo): message: adding rules after symbols not supported

I'm probably missing something, but libpotassco already provides `SmodelsConvert` for a more flexible/general conversion. `SmodelsOutput` is only meant for directly writing in smodels format - it does not maintain any additional state. 

Hence, not sure I understand the motivation for the new `SmodelsAdapter` class.

lib/c-api/src/control.cc
```diff
@@ -44,16 +44,16 @@ extern "C" auto clingo_control_new(clingo_lib_t *lib, clingo_string_t const *arg
...
-                         {"parse", CppClingo::Control::AppMode::parse},
-                         {"rewrite", CppClingo::Control::AppMode::rewrite},
-                         {"ground", CppClingo::Control::AppMode::ground},
-                         {"solve", CppClingo::Control::AppMode::solve},
-                     })),
-             "Run in {parse|rewrite|ground|solve} mode");
+            ("mode", parse(parse_mode),
+             "Run in mode\n"
+             "      %A: <mode {parse|rewrite|default}>\n"
+             "        parse   : Stop processing after parsing\n"
```
BenKaufmann: What happened here? In [commit](https://github.com/potassco/clingo/pull/599/changes/e268ccc4cf252c5c93f8b971c6ceb3e614858231), the help output was (imo) improved but now its back to the old (worse) one.

lib/c-api/src/main.cc
```diff
@@ -173,38 +167,54 @@ class ClingoApp : public Clasp::Cli::ClaspAppBase {
...
         root.add(group_basic);
         app_.register_options(root);
     }
 
     void validateOptions(const Potassco::ProgramOptions::OptionContext &root,
                          const Potassco::ProgramOptions::ParsedOptions &parsed) override {
         BaseType::validateOptions(root, parsed);
+        opts_.validate_options(parsed);
+        if (clasp_mode_ && (opts_.backend_type() != BackendType::clasp || opts_.mode() != AppMode::solve)) {
+            throw std::invalid_argument("incompatible values for '--convert' and '--mode'");
```
BenKaufmann: What happened here and why? In [commit](https://github.com/potassco/clingo/pull/599/changes/e268ccc4cf252c5c93f8b971c6ceb3e614858231), the exception type was changed to `Potassco::ProgramOptions::Error` to get a more unified output but now it is back to the old type.

rkaminsk: I messed up the merge. I'll fix this.

lib/c-api/src/opts.hh
```diff
@@ -237,7 +296,40 @@ class ClingoOptions {
...
+        if (it == items.end()) {
+            return false;
+        }
+        solver_opts_.mode = it->first;
+        return true;
+    }
+
+    void validate_options([[maybe_unused]] Potassco::ProgramOptions::ParsedOptions const &parsed) {
+        if (ground_mode_) {
+            if (backend_type() != Control::BackendType::clasp) {
```
BenKaufmann: See previous comment regarding exception type.

rkaminsk:
> > There also seems to be a bug with smodels conversion:
> > ```
> > 
> > ❯ echo '{x; y} = 1.' | clingo --convert=smodels
> > 
> > *** ERROR: (clingo): SmodelsOutput &Potassco::SmodelsOutput::startRule(SmodelsType):339: Precondition 'sec_ == 0 || rt == SmodelsType::end || rt >= SmodelsType::clasp_increment' failed.
> > 
> > *** Info : (clingo): message: adding rules after symbols not supported
> > ```
> 
> I'm probably missing something, but libpotassco already provides `SmodelsConvert` for a more flexible/general conversion. `SmodelsOutput` is only meant for directly writing in smodels format - it does not maintain any additional state.
> 
> Hence, not sure I understand the motivation for the new `SmodelsAdapter` class.

I just quickly pushed something to make it work. We'll switch to SmodelsConvert of course.

lib/control/src/solver.cc
```diff
@@ -1068,23 +1117,129 @@ auto SymbolTable::output(CppClingo::Symbol const &sym) -> State & {
...
-auto Solver::make_output_(SymbolStore &store, AppMode mode) -> UOutputStm {
-    switch (mode) {
-        case AppMode::solve: {
-            auto backend = std::make_unique<ProgramBackendImpl>(*clasp_->asp(), terms_);
-            theory_ = std::make_unique<Output::TheoryData>(store, *backend);
-            backend_ = std::move(backend);
-            return Output::make_backend_output(store, *backend_, *theory_);
+namespace {
+
+class SmodelsAdapter : public Potassco::AbstractProgram {
```
BenKaufmann: Can we use `Potassco::SmodelsConvert` instead?

lib/c-api/src/main.cc
BenKaufmann: Personal opinion: I strongly prefer rebasing when it comes to updating feature branches.

rkaminsk: I did not want to force push because now two people were at work here. Given the number of commits, I'll also squash merge this. In the end it's all my fault, I should just have done that pull before starting to hack away... Sorry for the extra work I caused.

rkaminsk:
I also added some test cases. I hope this covers every thing now. Let me know whether you still spot things that should be changed/improved.

lib/c-api/src/opts.hh
```diff
@@ -171,6 +172,54 @@ class ClingoOptions {
...
+            namespace Parse = Potassco::Parse;
+            using namespace std::string_view_literals;
+
+            if (auto key = "text"sv; Parse::eqIgnoreCase(str, key)) {
+                ground_mode_ = true;
+                return true;
+            }
+
+            auto backend_type = Control::BackendType{};
+            auto items = std::array{
```
BenKaufmann: Consider making the array `static constexpr`

lib/c-api/src/opts.hh
```diff
@@ -237,7 +296,40 @@ class ClingoOptions {
         }
     }
 
+    auto init_app_mode(std::string_view str) -> bool {
+        namespace Parse = Potassco::Parse;
+        using namespace std::string_view_literals;
+
+        auto items = std::array{
```
BenKaufmann: Consider making the array `static constexpr`

lib/c-api/src/opts.hh
```diff
@@ -171,6 +172,54 @@ class ClingoOptions {
...
+        auto parse_convert = [this](std::string_view str) {
+            namespace Parse = Potassco::Parse;
+            using namespace std::string_view_literals;
+
+            if (auto key = "text"sv; Parse::eqIgnoreCase(str, key)) {
+                ground_mode_ = true;
+                return true;
+            }
+
+            auto backend_type = Control::BackendType{};
```
BenKaufmann: Nitpick: You can move the definition of `backend_type` to line 197 so that it is directly initialized with the correct value:
```
auto backend_type = it->first
```

lib/c-api/src/opts.hh
```diff
@@ -237,7 +296,40 @@ class ClingoOptions {
...
+        return true;
+    }
+
+    void validate_options([[maybe_unused]] Potassco::ProgramOptions::ParsedOptions const &parsed) {
+        if (ground_mode_) {
+            if (backend_type() != Control::BackendType::clasp) {
+                throw Potassco::ProgramOptions::Error("incompatible values for options '--convert' and '--mode'");
+            }
+            mode() = Control::AppMode::ground;
+        } else {
```
BenKaufmann: In general, I'm not a huge fan of this validation approach since its result is order-dependent.
E.g. `--convert=reify --mode=rewrite` will trigger an error, while `--mode=rewrite --convert=reify` won't (AppMode is set to `solve` in the convert parser)

One possible fix for this would be dropping the `solver_opts_.mode = Control::AppMode::solve;` in `parse_convert`.

lib/control/src/solver.cc
```diff
@@ -245,16 +229,83 @@ class AbstractProgramBackendImpl : public ProgramBackend, public TheoryBackend {
...
+        if (auto lit = next_atom_++; std::cmp_less_equal(lit, prg_lit_max)) {
+            return static_cast<prg_lit_t>(lit);
+        }
+        throw std::range_error("number of literals exhausted");
+    }
+
+    auto do_fact_lit() -> std::optional<prg_lit_t> override {
+        if (fact_ == 0) {
+            fact_ = do_next_lit();
+            program().rule(Potassco::HeadType::disjunctive, Potassco::AtomSpan{&fact_, 1}, {});
```
BenKaufmann: Nitpick: you could use `Potassco::toSpan(fact_)` here.

lib/util/include/clingo/util/print.hh
BenKaufmann:
```suggestion
    //! Construct the buffer with an optional file handle.
```

lib/c-api/src/opts.hh
rkaminsk: I missed that the convert parser sets the mode to solve. The whole idea of my change was that the convert parser does not touch the mode.

rkaminsk: The fix is easy: we just don't set the mode in the convert parser.

lib/control/include/clingo/control/solver.hh
```diff
@@ -844,9 +866,11 @@ class Solver : public BaseView {
     TermBaseMap terms_;
     Clasp::ClaspFacade *clasp_;
     ClingoConfig config_;
-    Util::OutputBuffer buf_;
+    Util::OutputStream buf_;
     UProgramBackend backend_;
     std::unique_ptr<Output::TheoryData> theory_;
+    std::unique_ptr<Potassco::AbstractProgram> output_program_;
+    std::unique_ptr<Potassco::AbstractProgram> program_;
```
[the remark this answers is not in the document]
rkaminsk: Added `#include <potassco/basic_types.h>` even though it comes transitively.

lib/control/src/solver.cc
```diff
@@ -1167,11 +1246,11 @@ void Solver::main() {
...
     } else {
         if (opts_.mode == AppMode::parse) {
-            output_unprocessed_program(std::cout);
+            output_unprocessed_program(buf_);
             return;
         }
         if (opts_.mode == AppMode::rewrite) {
-            output_program(std::cout);
+            output_program(buf_);
             return;
```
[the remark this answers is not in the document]
rkaminsk: The two involved functions flush once output is complete. It also fits logically that they do this.

lib/util/include/clingo/util/print.hh
```diff
@@ -196,6 +197,44 @@ class OutputBuffer {
...
+    class Buffer : public std::streambuf {
+      public:
+        Buffer(FILE *out) : out_{out} {}
+        auto buffer() -> OutputBuffer & { return out_; }
+
+      protected:
+        auto overflow(int ch) -> int override {
+            if (ch != traits_type::eof()) {
+                out_.push_back(static_cast<char>(ch));
+                return ch;
```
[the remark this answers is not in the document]
rkaminsk: Done.

lib/util/include/clingo/util/print.hh
```diff
@@ -196,6 +197,44 @@ class OutputBuffer {
     FILE *out_;
 };
 
+//! Helper to use OutputBuffer with iostreams.
+class OutputStream : public std::ostream {
+  public:
+    // Construct the stream with an optional file handle.
+    OutputStream(FILE *out = nullptr) : std::ostream{&buf_}, buf_{out} {}
```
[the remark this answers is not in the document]
rkaminsk: I actually thought about it when writing the code and I always find this dance annoying. But why not...

lib/c-api/src/opts.hh
BenKaufmann, line 321: Nitpick: Consider removing the duplication:
```cpp
if (backend_type() != Control::BackendType::clasp &&
            (ground_mode_ || mode() == Control::AppMode::rewrite || mode() == Control::AppMode::parse)) {
            throw Potassco::ProgramOptions::Error("incompatible values for options '--convert' and '--mode'");
}
if (ground_mode_) {
    mode() = Control::AppMode::ground;
}
```

@rkaminsk  Sorry, it seems I messed up and forgot to press the "comment" button on my last round.

lib/util/include/clingo/util/print.hh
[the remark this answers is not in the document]
rkaminsk: Setting the rdbuf later causes segfaults. So now I just manage the buffer externally. Since the STL provides streams that manage their own buffers, they must have solved this somehow. Did I already complain about the iostream library? That's hereby done!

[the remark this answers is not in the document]
BenKaufmann: > Setting the rdbuf later causes segfaults.

Weird. Can you tell me where/when this segfaults? I looked through the standard and also tested with various compilers and from what I can see, the following both should be legal and ok:
```cpp
OutputStream(FILE *out = nullptr) : std::ostream{nullptr}, buf_{out} {
  assert(this->bad());  // no output possible if initialized with sb = nullptr
  this->rdbuf(&buf_);  // restores stream state  
  assert(this->good());
}
```
```cpp
OutputStream(FILE *out = nullptr) : std::ostream{nullptr}, buf_{out} {
  assert(this->bad());  // no output possible if initialized with sb = nullptr
  this->init(&buf_);  // from my understanding, calling init twice is allowed
  assert(this->good());
}
```

Anyhow, this is mostly for my curiosity. I'm otherwise fine with the new code.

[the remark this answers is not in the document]
rkaminsk: I used the default constructor and did not pass a nullptr. Maybe this was the issue...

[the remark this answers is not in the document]
BenKaufmann: > I used the default constructor and did not pass a nullptr. Maybe this was the issue...

That's indeed the issue. The standard doesn't even mandate a default ctor (and MSVC does not provide one). With `libc++`, the default ctor _intentially does not initialize_ and any call to a member function other than `init()` (i.e. including `rdbuf()`) has UB.

In short: with the default ctor, only the following would work with clang and libc++:
```cpp
OutputStream(FILE *out = nullptr) : std::ostream{}, buf_{out} {
  this->init(&buf_); 
}
```
However, it wouldn't be portable and one should instead use `std::ostream{nullptr}`.

lib/control/src/solver.cc
```diff
@@ -245,16 +229,83 @@ class AbstractProgramBackendImpl : public ProgramBackend, public TheoryBackend {
...
+
+    virtual void do_term_id(Symbol sym, prg_id_t id) = 0;
+
+    auto as_atoms_(PrgLitSpan lits) -> Potassco::AtomSpan {
+        atoms_.clear();
+        for (auto const &lit : lits) {
+            assert(lit > 0);
+            atoms_.push_back(lit);
+        }
+        return Potassco::AtomSpan{atoms_};
```
BenKaufmann: Nitpick: Consider simplifying to `return atoms_`.

lib/control/src/solver.cc
```diff
@@ -245,16 +229,83 @@ class AbstractProgramBackendImpl : public ProgramBackend, public TheoryBackend {
...
+    }
+
+    auto as_wlits_(WeightedPrgLitSpan wlits) -> Potassco::WeightLitSpan {
+        wlits_.clear();
+        for (auto const &[lit, weight] : wlits) {
+            assert(lit != 0);
+            wlits_.emplace_back(lit, weight);
+        }
+        return Potassco::WeightLitSpan{wlits_};
```
BenKaufmann: Nitpick: Consider simplifying to `return wlits_`.

BenKaufmann (approved):

lib/util/include/clingo/util/print.hh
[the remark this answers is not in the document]
rkaminsk: On a second thought I could have also used private inheritance to change the initialization order. This would have been the clean hack to get everything in order providing the same interface. Anyway, the current solution works.

rkaminsk:
@BenKaufmann What do you think about the last commit? I have a tendency to prefer this one because it nicely bundles stream and buffer.

BenKaufmann:
> @BenKaufmann What do you think about the last commit? I have a tendency to prefer this one because it nicely bundles stream and buffer.

Looks good to me 👍

PR #600, opened 2026-02-09
Update RE2C version requirement to 3.0

haampie:
The short `--header` flag is RE2C v3+ only: https://github.com/skvadrik/re2c/commit/886b73af33335644f0d3909b54c12c6d6f478356


rkaminsk:
Thanks. Makes sense for clingo 6. We require a lot of other recent tools, anyway.

PR #604, opened 2026-02-20
Statistics

BenKaufmann:
@rkaminsk This is a draft PR just to gather some feedback from you.

Key questions:

- What's your opinion on making "user_accu"/"user_stats" part of the public C-API? (commit: 52fbbbc627b0e695f79d23bb02ada01954a95be9)
- What's your opinion on the new try_at function? (commit: 094e8ca988f271fa40e9806f7cc5b3271f6720bc). I also considered adding a second out parameter to `clingo_stats_map_has_subkey`, which seems more natural but would make certain call wrappers more complicated.
- What do you think about the reworked Python stats API (commit: 1b40d596ef0f9ad61d764305f12d9522dae80089)?
  - I added support for `__eq__` so that tests could still use asserts like `assert ctl.stats["summary"]["lower"] == [3.0]`
  - Do we need `keys()`, `values()`, `items()` for map?
  - Should (Const)Stats also implement `__iter__`?
  - Should we additionally add `__getattr__`?
- What do you think about an "item" based map interface, i.e. having `clingo_stats_map_item()` instead of `clingo_stats_map_subkey_name` + `clingo_stats_map_at`?

### DIFF
--- a/lib/c-api/include/clingo/stats.h
+++ b/lib/c-api/include/clingo/stats.h
@@ -58,6 +58,11 @@ typedef int clingo_stats_type_t;
 //! Handle for the solver stats.
 typedef struct clingo_statistic clingo_stats_t;
 
+//! Well-known key name intended to be used for user-specific step statistics.
+CLINGO_VISIBILITY_DEFAULT extern clingo_string_t const clingo_user_stats_step;
+//! Well-known key name intended to be used for user-specific accu statistics.
+CLINGO_VISIBILITY_DEFAULT extern clingo_string_t const clingo_user_stats_accu;
+
 //! Get the root key of the stats.
 //!
 //! @param[in] stats the target stats
--- a/lib/c-api/include/clingo/stats.h
+++ b/lib/c-api/include/clingo/stats.h
@@ -166,6 +166,20 @@ CLINGO_VISIBILITY_DEFAULT bool clingo_stats_map_subkey_name(clingo_stats_t const
 //! @return whether the call was successful
 CLINGO_VISIBILITY_DEFAULT bool clingo_stats_map_at(clingo_stats_t const *stats, uint64_t key, char const *name,
                                                    size_t size, uint64_t *subkey);
+
+//! Try to get the subkey under the given name or return the given default key if it does not exist.
+//!
+//! @pre The @link clingo_stats_type() type@endlink of the entry must be @ref ::clingo_stats_type_map.
+//! @param[in] stats the target stats
+//! @param[in] key the key
+//! @param[in] name the name to look up the subkey
+//! @param[in] size the size of the name
+//! @param[in] defkey the key to return if name is not found
+//! @param[out] subkey the resulting subkey
+//! @return whether the call was successful
+CLINGO_VISIBILITY_DEFAULT bool clingo_stats_map_try_at(clingo_stats_t const *stats, uint64_t key, char const *name,
+                                                       size_t size, uint64_t defkey, uint64_t *subkey);
+
 //! Add a subkey with the given name.
 //!
 //! @pre The @link clingo_stats_type() type@endlink of the entry must be @ref ::clingo_stats_type_map.
--- a/lib/c-api/src/stats.cc
+++ b/lib/c-api/src/stats.cc
@@ -161,6 +161,19 @@ extern "C" auto clingo_stats_map_at(clingo_stats_t const *stats, uint64_t key, c
     CLINGO_CATCH;
 }
 
+extern "C" auto clingo_stats_map_try_at(clingo_stats_t const *stats, uint64_t key, char const *name, size_t size,
+                                        uint64_t defkey, uint64_t *subkey) -> bool {
+    CLINGO_TRY {
+        if (stats == nullptr || (name == nullptr && size > 0) || subkey == nullptr) {
+            return fail_arguments();
+        }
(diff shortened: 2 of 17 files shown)


rkaminsk:
> @rkaminsk This is a draft PR just to gather some feedback from you.
> 
> Key questions:
> 
>     * What's your opinion on making "user_accu"/"user_stats" part of the public C-API? (commit: [52fbbbc](https://github.com/potassco/clingo/commit/52fbbbc627b0e695f79d23bb02ada01954a95be9))

Fine with me. I think that the C++ API should provide a ready to use string view for this.

>     * What's your opinion on the new try_at function? (commit: [094e8ca](https://github.com/potassco/clingo/commit/094e8ca988f271fa40e9806f7cc5b3271f6720bc)). I also considered adding a second out parameter to `clingo_stats_map_has_subkey`, which seems more natural but would make certain call wrappers more complicated.

I'll have closer look later. I suppose this was done to reduce the number of API calls for some of the wrapper functions.

>     * What do you think about the reworked Python stats API (commit: [1b40d59](https://github.com/potassco/clingo/commit/1b40d596ef0f9ad61d764305f12d9522dae80089))?

It makes sense to split into Stats/ConstStats. 

>       * I added support for `__eq__` so that tests could still use asserts like `assert ctl.stats["summary"]["lower"] == [3.0]`

Not sure. There is probably no need for this beyond the unit tests.

>       * Do we need `keys()`, `values()`, `items()` for map?
>       * Should (Const)Stats also implement `__iter__`?

I generally try to implement all methods of an ABC:

- https://docs.python.org/3/library/collections.abc.html#collections.abc.Mapping
- https://docs.python.org/3/library/collections.abc.html#collections.abc.MutableMapping
- https://docs.python.org/3/library/collections.abc.html#collections.abc.Sequence
- https://docs.python.org/3/library/collections.abc.html#collections.abc.MutableSequence

Of course, only if it makes sense for the statistics.

>       * Should we additionally add `__getattr__`?

I would not implement two ways to access the statistics.

>     * What do you think about an "item" based map interface, i.e. having `clingo_stats_map_item()` instead of `clingo_stats_map_subkey_name` + `clingo_stats_map_at`?

We can choose whatever makes the C/C++/Python API easier to implement.

BenKaufmann:
> > ```
> >   * Do we need `keys()`, `values()`, `items()` for map?
> >   * Should (Const)Stats also implement `__iter__`?
> > ```
> 
> I generally try to implement all methods of an ABC:

~Ok. Adding `keys()`,  `values()`, `items()` for map was easy. However, I'm not sure how to implement `__Iter__` for (Const)Stats since the iterator (return) type depends on the type of the stats object; string (key) for maps and ConstStats (element)  for arrays. Am I missing something?~

**EDIT:** Sorry for the noise. Figured that I could just use `py::iterator` as the common type.

rkaminsk:
> > > ```
> > >   * Do we need `keys()`, `values()`, `items()` for map?
> > >   * Should (Const)Stats also implement `__iter__`?
> > > ```
> > 
> > 
> > I generally try to implement all methods of an ABC:
> 
> ~Ok. Adding `keys()`, `values()`, `items()` for map was easy. However, I'm not sure how to implement `__Iter__` for (Const)Stats since the iterator (return) type depends on the type of the stats object; string (key) for maps and ConstStats (element) for arrays. Am I missing something?~
> 
> **EDIT:** Sorry for the noise. Figured that I could just use `py::iterator` as the common type.

I'll try to have a closer look next week. I find it important that the types work well in IDEs and tools like mypy/ruff. If the iterators are implemented in the StatsMap and StatsArray subclasses, then there should be no problem.

rkaminsk:
I tweaked a bit to get proper type hints generated. I think we should kick out the `__eq__` and `__float__` methods. I don't think that casting to float is nicer than using `.value`.

I also don't think `__eq__` is useful from a library perspective because we just don't compare stats in practice. Also in Python one has to implement `__ne__` as well. There is no automagic implementation of it.

At the moment, we have keys/values/items for views but not for the mutable derived classes. I think, they should also be implemented for the derived classes to return iterators to the mutable versions.

BenKaufmann:
> I tweaked a bit to get proper type hints generated. 

Thanks :+1:  Will add this to the final PR.

> I think we should kick out the `__eq__` and `__float__` methods. I don't think that casting to float is nicer than using `.value`.
> 

Ok. Will drop.  However, this means that I'll have to adjust some other tests (make them slightly uglier).

> At the moment, we have keys/values/items for views but not for the mutable derived classes. I think, they should also be implemented for the derived classes to return iterators to the mutable versions.

Ok. Not sure I agree, but will add the corresponding versions.

rkaminsk:
> > I tweaked a bit to get proper type hints generated.
> 
> Thanks 👍 Will add this to the final PR.
> 
> > I think we should kick out the `__eq__` and `__float__` methods. I don't think that casting to float is nicer than using `.value`.
> 
> Ok. Will drop. However, this means that I'll have to adjust some other tests (make them slightly uglier).

I think that's fine.

> > At the moment, we have keys/values/items for views but not for the mutable derived classes. I think, they should also be implemented for the derived classes to return iterators to the mutable versions.
> 
> Ok. Not sure I agree, but will add the corresponding versions.

I can imagine that there are use cases for this. Imagine the statistics have an array. Then we can write something like

```
for x in stats["array"].array():
  x.value += 10
```

without having to use the index based access. In general, the update method should be used though.

PR #605, opened 2026-02-25
Update clasp and rework python stats API

### LINKED ISSUE #540: Clingo on Android 11+ with python support
This is an update on https://github.com/potassco/clingo/issues/475, which explained how to build clingo for Android under Termux. The steps described in that issue fail to make the clingo API accessible from Python programs. I thought I would share the updated steps in case someone other than me needs to build clingo on Android/Termux. If there is enough interest, we might look into releasing a clingo package for Android/Termux and include a suitable patch there.

**Symptom**
Running the following command gives error:
```
% python -c "from clingo import Control"
ImportError: dlopen failed: cannot locate symbol "_Py_NoneStruct" referenced by "/data/data/com.termux/files/usr/lib/python3.12/site-packages/clingo/_clingo.cpython-312.so"...
```

**Solution**
The problem comes from the fact that the Python package for Termux does not currently export its symbols globally. While we wait for the issue to be addressed by the package owner, the following updated steps provide a workaround.

*Step 1:* build dmalloc with `-fPIC`
Download dmalloc from https://dmalloc.com and extract it, then cd to the folder that was created. Build it with:
```
env CFLAGS=-fPIC LDFLAGS=-fPIC ./configure --prefix=/data/data/com.termux/files/usr/
make install
```
Note that the part `CFLAGS=-fPIC LDFLAGS=-fPIC` is critical.

*Step 2:* download the clingo tarball, extract it, and cd to the folder that was created. Then, build clingo and explicitly link against dmalloc **and the python library**
```
cmake -H. -Bbuild -DCMAKE_BUILD_TYPE=Release -DCMAKE_EXE_LINKER_FLAGS="/data/data/com.termux/files/usr/lib/libdmalloc.a" -DCMAKE_EXE_LINKER_FLAGS_RELEASE="/data/data/com.termux/files/usr/lib/libdmalloc.a" -DCMAKE_SHARED_LINKER_FLAGS="/data/data/com.termux/files/usr/lib/libdmalloc.a -L/data/data/com.termux/files/usr/lib -lpython3" -DCMAKE_INSTALL_PREFIX="/data/data/com.termux/files/usr/"

cmake --build build --target install
```
Note that the command above assumes the use of Python 3.x. I have not tested Python 2.x.

That's it. `python -c "from clingo import Control` should work properly now. For a more thorough test, run the chemistry example found in the clingo distribution:
```
cd examples/clingo/chemistry/
python app.py
```
If the program runs without errors, clingo's API is accessible from Python programs.

BenKaufmann:
#### extend stats C-API
* make key names for user stats part of the public API
* add (optional) subkey parameter to `clingo_stats_map_has_subkey` to
  enable more efficient "get-if-found" queries

#### rework python stats API
* add view (read-only) types for Stats, StatsArray, and StatsMap
  and expose `Control.stats` as a read-only and lazy StatsView
  instead of an eagerly populated dictionary

### DIFF
--- a/lib/c-api/include/clingo/stats.h
+++ b/lib/c-api/include/clingo/stats.h
@@ -58,6 +58,11 @@ typedef int clingo_stats_type_t;
 //! Handle for the solver stats.
 typedef struct clingo_statistic clingo_stats_t;
 
+//! Well-known key name intended to be used for user-specific step statistics.
+CLINGO_VISIBILITY_DEFAULT extern clingo_string_t const clingo_user_stats_step;
+//! Well-known key name intended to be used for user-specific accu statistics.
+CLINGO_VISIBILITY_DEFAULT extern clingo_string_t const clingo_user_stats_accu;
+
 //! Get the root key of the stats.
 //!
 //! @param[in] stats the target stats
@@ -135,10 +140,12 @@ CLINGO_VISIBILITY_DEFAULT bool clingo_stats_map_size(clingo_stats_t const *stats
 //! @param[in] key the key
 //! @param[in] name name of the subkey
 //! @param[in] size the size of the name
+//! @param[out] subkey resulting subkey if @p result is true
 //! @param[out] result true if the map has a subkey with the given name
 //! @return whether the call was successful
+//! @note @p subkey is optional and can be set to NULL to only check for existence
 CLINGO_VISIBILITY_DEFAULT bool clingo_stats_map_has_subkey(clingo_stats_t const *stats, uint64_t key, char const *name,
-                                                           size_t size, bool *result);
+                                                           size_t size, uint64_t *subkey, bool *result);
 //! Get the name associated with the offset-th subkey.
 //!
 //! @pre The @link clingo_stats_type() type@endlink of the entry must be @ref ::clingo_stats_type_map.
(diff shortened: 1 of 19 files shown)


rkaminsk:
I like it very much.

For discussion: You split mutable/immutable stats into Stats and StatsView in the python API. Maybe we should also follow this naming convention in the C++ classes? Now that you suggested it, I like the view variant better. We don't have to do this in this PR and can do it separately afterward also addressing the config.

BenKaufmann:
> For discussion: You split mutable/immutable stats into Stats and StatsView in the python API. Maybe we should also follow this naming convention in the C++ classes? 

As one can see from the implementation names, I was somehow inconsistent/confused and copied the X vs. ConstX naming scheme. For the API, I'd prefer X/XView, though. Given that in C++ the implementation **is** the API, renaming makes sense to me. Let me know if you'd like me to do it as part of this PR.

As a more general question: Have you thought about implementing the python API in terms of the C++ API (or at least sharing some code between them)?

rkaminsk:
> Let me know if you'd like me to do it as part of this PR.

Let's do a separate one to get this one done.

> As a more general question: Have you thought about implementing the python API in terms of the C++ API (or at least sharing some code between them)?

I thought of it at times. But then there are subtle differences in underlying C++ helper classes of the Python API and the classes of the C++ API; especially, when it comes to lifetime management. So I decided to keep both separate. Maybe I'll rethink this when switching to nanobind.

rkaminsk (approved):
I think that this is ready to merge.

PR #607, opened 2026-02-26
Refactor/const view

rkaminsk:


lib/python-api/src/stats.cc
```diff
@@ -10,31 +10,31 @@ class StatsIterBase {
...
         return {name, value};
     }
-    [[nodiscard]] auto const_array_item() const -> ConstStats { return stats_.as<ConstStatsArray>().get(pos_); }
+    [[nodiscard]] auto const_array_item() const -> StatsView { return stats_.as<StatsArrayView>().get(pos_); }
     [[nodiscard]] auto array_item() const -> Stats { return stats_.as<StatsArray>().get(pos_); }
 
   private:
     StatsBase stats_;
-    std::size_t pos_;
+    std::size_t pos_ = 0;
```
BenKaufmann, line 37: FYI: That's how I wrote it in the first place but then one of the "flow analyzers" in my IDE (**wrongly**) complained about "Usage of non-initialized class field 'pos_' when called from function 'nestify'". That's why I moved it to the initializer list.

lib/cxx-api/tests/stats.cc
```diff
@@ -62,7 +62,11 @@ TEST_CASE_METHOD(Fixture, "stats solve", "[cxx][stats][solve]") {
     REQUIRE(models == MV{{"a"}, {"b"}, {"c"}, {"d"}});
     auto stats = ctl.stats();
     REQUIRE(*stats["solving"]["solvers"]["choices"] > 0);
+#ifndef __EMSCRIPTEN__
```
BenKaufmann, line 65: Sorry about that :expressionless:
We could alternatively just test for "total" time, which is also present with __EMSCRIPTEN__.

BenKaufmann (approved):

lib/python-api/src/stats.cc
rkaminsk: The pre-commit linter complained about it and I generally do it like this in the rest of clingo. Sometimes linters can be annoying...

lib/cxx-api/tests/stats.cc
rkaminsk: I think, we can keep it like this. Even if the test breaks in the future, it will be easy to fix.

PR #610, opened 2026-03-13
Flush stdout if a custom model printer has been registered.

rkaminsk:
Output might get tangled if a model printer is registered in Python because python manages its own output buffer. I pushed a preliminary fix. @BenKaufmann what do you think about this? Should we flush the output in general or should we just do it for the Python API? For the C/C++ APIs there should be no issues because here we do not have separate buffers.

### DIFF
--- a/lib/python-api/src/app.cc
+++ b/lib/python-api/src/app.cc
@@ -1,3 +1,4 @@
+#include <iostream>
 #include <utility>
 
 #include "app.hh"
@@ -138,7 +139,11 @@ class App : public reference_keeper<App> {
                              void *data) -> bool {
         auto &app = *static_cast<App *>(data);
         CLINGO_TRY {
+            auto acquire = py::gil_scoped_acquire{};
+            // NOTE: Python seems to directly write large buffers
+            std::cout.flush();
             app.print_model(Model{model}, [printer, printer_data]() { handle_error(printer(printer_data)); });
+            py::module_::import("sys").attr("stdout").attr("flush")();
         }
         CLINGO_CATCH;
     }


rkaminsk:
With the last force push, this is handle on the Python side only:

```c++
// ensure OS buffer is flushed
std::cout.flush()
app.print_model(Model{model}, [printer, printer_data]() { handle_error(printer(printer_data)); });
// ensure Python buffer is flushed
py::module_::import("sys").attr("stdout").attr("flush")();
```

The two flushes are important because it seems like python directly writes large buffers without flushing previously written smaller buffers leading to tangled up output.

lib/python-api/src/app.cc
```diff
@@ -138,7 +139,11 @@ class App : public reference_keeper<App> {
                              void *data) -> bool {
         auto &app = *static_cast<App *>(data);
         CLINGO_TRY {
+            auto acquire = py::gil_scoped_acquire{};
+            // NOTE: Python seems to directly write large buffers
+            std::cout.flush();
```
BenKaufmann, line 144: While I understand the motivation for this flush here, I think it hides an actual issue in the solver. That is, imo the clasp component should ensure that all its output buffers are flushed before it calls back into user code. From what I can see, clasp's TextOutput currently only relies on the fact that (stdout) output is typically flushed on '\n' but it does not explicitly flush the sink (typically stdout) before/after printing a model. I'll address this in a separate update. 

Of course, if other (clingo) code also might write to stdout/stderr, we should keep the flush here.

lib/python-api/src/app.cc
```diff
@@ -138,7 +139,11 @@ class App : public reference_keeper<App> {
...
             app.print_model(Model{model}, [printer, printer_data]() { handle_error(printer(printer_data)); });
+            py::module_::import("sys").attr("stdout").attr("flush")();
```
BenKaufmann, line 146: Yes, flushing and thereby ensuring that all user output is written at this point makes sense to me. 

Btw: Do you have an example/test demonstrating the original issue?

BenKaufmann (approved):

lib/python-api/src/app.cc
rkaminsk: Yes, for the Python API we have to do this flush to play nice with the rest of clingo's output. Here is the example. You can play with the commented out flushes. The issue manifests if you redirect into a file:

```bash
echo '{a;b}.' | python test.py --order 0 > test.txt
```

```python
from typing import Callable, Sequence
import sys
from clingo.app import App, AppOptions, Flag, clingo_main
from clingo.solve import Model
from clingo.symbol import Symbol

Parts = Sequence[Sequence[tuple[str, Sequence[Symbol]]]]


class MyApp(App):
    def __init__(self) -> None:
        super().__init__("my-app", "1.0.0")
        self._order = Flag()

    def print_model(self, model: Model, default_printer: Callable[[], None]) -> None:
        # sys.stdout.flush()
        if self._order.value:
            print(" ".join(str(sym) for sym in sorted(model.symbols(shown=True))))
        else:
            default_printer()
        # sys.stdout.flush()

    def register_options(self, options: AppOptions) -> None:
        options.add_flag(
            "MyApp", "order", "Print atoms in models in order.", self._order
        )

    def main(self, control: Control, files: Sequence[str]) -> None:
        control.parse_files(files)
        control.main()


with Library() as lib:
    sys.exit(clingo_main(lib, sys.argv[1:], MyApp()))
```

lib/python-api/src/app.cc
rkaminsk: I am not 100% sure whether there is an issue in clasp because this should not manifest when the standard printing mechanisms are used. Which is a reasonable assumption or not?

lib/python-api/src/app.cc
BenKaufmann: Thanks. I'll have a look ASAP.

> The issue manifests if you redirect into a file:

This makes sense given that "flush on newline" (line buffering) typically only happens if stdout is not redirected. So independent of this PR, I guess I should still investigate whether clasp' s TextOutput behaves nicely when output is redirected.

jorgefandinno:
If I remember correctly `c` functions `fwrite` and `write` also may use different buffers in some systems. So, this issue may have happened in the c-api for some systems.

lib/python-api/src/app.cc
BenKaufmann: Looked at the code and tested with clasp. You are right: no additional flush in clasp needed.

rkaminsk:
> If I remember correctly `c` functions `fwrite` and `write` also may use different buffers in some systems. So, this issue may have happened in the c-api for some systems.

It may not be required by the standard but printf and fprintf to stdout not using the same buffer would be strange. In any case, users should stick to printf when using the C API. So I don't see a problem. And for other language bindings this has to be addressed on a case by case basis.

rkaminsk:
I checked the code and Python is using _Py_write_impl in cpython/Python/fileutils.c, which ultimatively uses POSIX write on Linux. So the initial flush is important to avoid garbled up output because printf is using it's own buffer. Simple example to reproduce:

```c++
#include <stdio.h>
#include <unistd.h>

#define N 4096

int main() {
    printf("small");
    char buf[N];
    for (int i = 0; i < N; ++i) {
        buf[i] = 'X';
    }
    write(STDOUT_FILENO, buf, N);
}
```

It most likely prints "small" at the very end.

PR #611, opened 2026-03-20
update clasp and use new update mode

BenKaufmann:
Use new (more efficient) clasp update mode when updating a solved program. With the new mode, we don't have to add an empty program when "re-solving" a solved program (e.g. only changing assumptions or reasoning modes), thereby avoiding the overhead incurred by preprocessing.

**NOTE:** We probably have to adjust this once https://github.com/potassco/clingo/pull/599 has been merged

### DIFF
--- a/lib/control/include/clingo/control/solver.hh
+++ b/lib/control/include/clingo/control/solver.hh
@@ -740,12 +740,14 @@ class Solver : public BaseView {
 
     //! Get the output buffer.
     //!
-    //! If the control object has been constructed without a null output FILE,
-    //! this buffer contains the output of the textoutput.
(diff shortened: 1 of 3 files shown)


rkaminsk (commented):
I have not tested yet but looking from far away this seems okay. The solve function proceeds as usual for everything but the clasp backend, which can do the right thing in both the ground+solve and resolve scenario.

BenKaufmann:
@rkaminsk 
Now that you have merged #599, let me update (rebase) and test this PR...

rkaminsk (commented):
This change should only affect how the code interfaces with clasp's facade. So if this is the correct way to do this using Clasp's API, then this should be fine. Ready to merge from my end.

PR #616, opened 2026-03-23
correctly track next_lit and next_term

rkaminsk:
This PR fixes ID tracking when parsing/forwarding ASPIF programs so that subsequent grounding steps allocate fresh literal/term IDs without collisions, and adds a regression test for incremental ASPIF conversion:

- Track maximum literal by updating internal counters whenever existing atoms/literals are encountered (rules, assumptions, outputs, theory conditions, etc.).
- Track maximum term id for output terms so IDs remain consistent across parse + later grounding.
- Add a Python test covering an incremental ASPIF parse followed by additional grounding and verifying the produced ASPIF.

### DIFF
--- a/lib/control/src/solver.cc
+++ b/lib/control/src/solver.cc
@@ -27,8 +27,13 @@ namespace {
 //! Implementation of the backend interface.
 //!
 //! This implementation forwards grounded statements to a
-//! Potassco::AbstractProgram. As such it can create neither atom ids nor term
-//! ids, which is the responsibility of derived classes.
+//! Potassco::AbstractProgram. The implementation leaves the management of term
+//! ids for show statements to derived classes.
+//!
+//! @note Maximum literal tracking is not required by all derived classes. For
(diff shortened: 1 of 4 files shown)


lib/control/src/solver.cc
```diff
@@ -229,25 +231,48 @@ class AbstractProgramBackendImpl : public ProgramBackend, public TheoryBackend {
...
+        return lit;
+    }
+
+    auto as_lit_(prg_lit_t lit) -> Potassco::Lit_t {
+        max_lit_ = std::max(max_lit_, std::abs(lit));
+        return lit;
+    }
+    auto as_lits_(PrgLitSpan body) -> Potassco::LitSpan {
+        for (auto lit : body) {
+            assert(lit != 0);
```
BenKaufmann: For consistency with `as_atom_`, I would suggest moving this assert to `as_lit_`

BenKaufmann (approved):

PR #617, opened 2026-03-24
use shared mutex for acessing/creating symbols

rkaminsk:


libgringo/src/symbol.cc
```diff
@@ -25,9 +25,11 @@
 #include <algorithm>
 #include <cstring>
 #include <gringo/hash_set.hh>
+#include <gringo/shared_mutex.hh>
 #include <gringo/symbol.hh>
 #include <iterator>
 #include <mutex>
+#include <shared_mutex>
```
BenKaufmann, line 32: Given that you are using this C++14-header, have you considered using  `std::shared_timed_mutex`  instead of the platform-specific implementations?
It's probably not as efficient (on Windows at least), though.

libgringo/gringo/shared_mutex.hh
```diff
@@ -0,0 +1,116 @@
+#pragma once
+
+#if defined(_WIN32) || defined(_WIN64)
```
BenKaufmann, line 3: I strongly agree with copilot regarding the macro naming.
Additionally, I'm not a fan of this kind of platform-handling in a header. More on this later...

libgringo/gringo/shared_mutex.hh
```diff
@@ -0,0 +1,116 @@
...
+#elif RWL_PLATFORM_POSIX
+        return pthread_rwlock_trywrlock(&rw_) == 0;
+#else
+        return mtx_.try_lock();
+#endif
+    }
+
+    void unlock() noexcept {
+#if RWL_PLATFORM_WINDOWS
+        ReleaseSRWLockExclusive(&srw_);
```
BenKaufmann, line 117: This looks broken to me. If the lock was taken in shared mode, I would assume that you'd also have to call `ReleaseSRWLockShared`

libgringo/gringo/shared_mutex.hh
```diff
@@ -0,0 +1,116 @@
...
+#endif
+
+#if !RWL_PLATFORM_WINDOWS && (defined(__unix__) || defined(__APPLE__) || defined(_POSIX_VERSION))
+#define RWL_PLATFORM_POSIX 1
+#else
+#define RWL_PLATFORM_POSIX 0
+#endif
+
+#if RWL_PLATFORM_WINDOWS
+#include <Windows.h>
```
BenKaufmann, line 22: Consider defining WIN32_LEAN_AND_MEAN and probably also NOMINMAX before including windows.h

**EDIT:**
> not enough arguments for function-like macro invocation 'max'

And there you have the motivation for using NOMINMAX or not including ugly windows.h :smile:

libgringo/gringo/shared_mutex.hh
```diff
@@ -0,0 +1,116 @@
...
+#elif RWL_PLATFORM_POSIX
+#include <pthread.h>
+#else
+#include <mutex>
+#endif
+
+namespace Gringo {
+
+class SharedMutex {
```
BenKaufmann, line 44: Personal preference: I would separate the interface from the platform-specific implementation.
I.e.
```cpp
#if defined(_WIN32) || defined(_WIN64)
#define GRINGO_PLATFORM_WINDOWS 1
struct SharedMutexType {
    void* lock; // SRWLock is pointer-sized - no need to include windows.h here
    bool read{false}; // SRWLock distinguishes between Shared/Exclusive unlock, but we only expose one function
};
#else
#define GRINGO_PLATFORM_WINDOWS 0
#endif

#if !GRINGO_PLATFORM_WINDOWS && (defined(__unix__) || defined(__APPLE__) || defined(_POSIX_VERSION))
#define GRINGO_PLATFORM_POSIX 1
#include <pthread.h>
using SharedMutexType = pthread_rwlock_t;
#else
#define GRINGO_PLATFORM_POSIX 0
#endif

#if !GRINGO_PLATFORM_WINDOWS && !GRINGO_PLATFORM_POSIX
#include <mutex>
using SharedMutexType = std::mutex;
#endif

namespace Gringo {
namespace Detail {
// Implemented in a (platform-specific) cpp file where we can include ugly headers like windows.h
void init(SharedMutexType& m);
void release(SharedMutexType& m);
void lock(SharedMutexType& m, bool exclusive);
bool try_lock(SharedMutexType&, bool exclusive);
// uses read member on Windows
void unlock(SharedMutexType&);
}

class SharedMutex {
  public:
    SharedMutex() noexcept { Detail::init(rw_); }
    ~SharedMutex() noexcept { Detail::release(rw_);}
    SharedMutex(const SharedMutex &) = delete;
    SharedMutex &operator=(const SharedMutex &) = delete;
    SharedMutex(SharedMutex &&) = delete;
    SharedMutex &operator=(SharedMutex &&) = delete;

    void lock_shared() noexcept { Detail::lock(rw_, false); }
    bool try_lock_shared() noexcept { return Detail::try_lock(rw_, false); }
    void unlock_shared() noexcept { Detail::unlock(rw_); }

    void lock() noexcept { Detail::lock(rw_, true); }
    bool try_lock() noexcept { return Detail::try_lock(rw_, true); }
    void unlock() noexcept { Detail::unlock(rw_); }

  private:
    SharedMutexType rw_{};
};

} // namespace Gringo
```
And then, either have one shared_mutex.cpp or shared_mutex_{win,posix,fallback} implementing the Detail functions.

rkaminsk: There is really no need to clean up too much here. This header is going to be used in a single rather small implementation file. Clingo has no need for a shared lock in other places. So we can keep it simple.

libgringo/gringo/shared_mutex.hh
rkaminsk: Tracking the state seems unnecessary. The only purpose might be to catch errors. Matching lock/unlock functions have to be used. On windows we have

    void unlock() noexcept {
        ReleaseSRWLockExclusive(&srw_);
    }

    void unlock_shared() noexcept {
        ReleaseSRWLockShared(&srw_);
    }

I checked the implementation of shared_lock and unique_lock, which call the expected functions.

BenKaufmann: My bad. For some reason, I missed the `unlock_shared` function.

libgringo/src/symbol.cc
rkaminsk: Since the code path is performance critical, I'd rather stick with the the lightweight locking primitives. I should however use the shared_timed_mutex as the fallback instead of the std::mutex.

rkaminsk: I looked at the implementation of std::shared_timed_mutex and at least on linux we could even just use it. The problem is that we support really old macos platforms (10.9) where it is not available. That's why I would simply keep the current implementation. It should be more or less the same as the stl implementation.

libgringo/src/symbol.cc
```diff
@@ -82,24 +84,28 @@ template <class T> struct UniqueConstruct {
...
+        {
+            auto lock = std::shared_lock<SharedMutex>{mutex_};
+            auto it = set_.find(x, hash);
+            if (it != set_.end()) {
+                return *it;
+            }
+        }
+        {
+            auto lock = std::unique_lock<SharedMutex>{mutex_};
+            return *set_.insert(T{std::forward<U>(x), hash}).first;
```
samutoljamo, line 97: `T{std::forward<U>(x)` allocates memory, I think you could move this outside `unique_lock` to have a slight perf benefit as allocation happens outside of mutex. Not sure if the compiler knows to optimize this over the mutex though.

libgringo/src/symbol.cc
```diff
@@ -82,24 +89,53 @@ template <class T> struct UniqueConstruct {
     using Set = hash_set<T, typename T::Hash, typename T::EqualTo>;
 
     template <class U> static auto construct(U &&x) {
-        // TODO: in C++17 this can use a read/write lock to not block reading threads
         size_t hash = typename T::Hash{}(x);
-        std::lock_guard<std::mutex> g(mutex_);
-        auto it = set_.find(x, hash);
-        if (it != set_.end()) {
-            return it->get();
+#if CLINGO_MAP_NUM_SHARDS > 1
```
BenKaufmann: Please consider reducing the logic duplication.
One trivial way to do so would be:
```cpp
auto& set = 
#if CLINGO_MAP_NUM_SHARDS > 1
	set_[hash % num_shards];
#else
	set_;
#endif
        {
            auto lock = std::shared_lock<SharedMutex>{mutex_};
            auto it = set.find(x, hash);
            if (it != set.end()) {
                return it->get();
            }
        }
        auto val = T{std::forward<U>(x), hash};
        {
            auto lock = std::unique_lock<SharedMutex>{mutex_};
            return set.insert(std::move(val)).first->get();
        }
```
Alternatively, you could have a simple "get_set" function.

libgringo/src/symbol.cc
```diff
@@ -82,24 +89,53 @@ template <class T> struct UniqueConstruct {
...
+            auto lock = std::unique_lock<SharedMutex>{mutex_};
+            return set_.insert(std::move(val)).first->get();
+        }
+#endif
     }
 
   private:
-    static Set set_;          // NOLINT
-    static std::mutex mutex_; // NOLINT
+#if CLINGO_MAP_NUM_SHARDS > 1
```
BenKaufmann: Is the distinction really needed?
Always using `static Set set_[num_shards]` seems fine to me. You'd even get a nice compile-time error if someone configures `CLINGO_MAP_NUM_SHARDS` to <= zero.

CMakeLists.txt
```diff
@@ -66,7 +66,9 @@ option(CLINGO_USE_LOCAL_CATCH
 ON)
 set(CLINGO_MAP_TYPE "hopscotch" CACHE STRING
 "Select hash table implementation. (sparse, hopscotch)")
-
+set(CLINGO_MAP_NUM_SHARDS "1" CACHE STRING
+"Select the number of shards for the global symbol table to improve concurrent \
```
BenKaufmann: Maybe consider "Set" instead of "Select". At least to me, the latter somehow suggests that there is a small/finite set of options.

libgringo/src/symbol.cc
rkaminsk: I was thinking about this too but admit being lazy because clingo 5 is reaching its end of life. Probably, one could even drop the if completely. The compiler should be able to handle the one element size array very well and detect that `hash % 1` is zero.

libgringo/src/symbol.cc
rkaminsk: I simplified it.

BenKaufmann (approved):

Issue #123, opened 2018-09-17
printing ast of 0-ary external functions


javier-romero:
Hi, 

Printing the ast of a 0-ary external function the parentheses are omitted, but they are necessary in clingo input files.
It would be nice if they were printed (from `__str__` or `__repr__` in Python).

I attach a simple example modifying clingo/examples/gringo/ast/visitor.lp ([visitor.txt](https://github.com/potassco/clingo/files/2388445/visitor.txt))
where running `gringo --text visitor.txt` it can be seen that `@foo()` is printed as `@foo.`

Thanks!
Javier

rkaminsk:
It would probably be best to extend the parser accept this. This maintains backward compatibility and makes documentation easier. Because an external function call is written as a functional term with a preceding `@` symbol. Parenthesis in functional terms are optional.

javier-romero:
And something similar about printing #const declarations.

Printing the ast for 
  `#const x=1.`
we obtain 
  `#const x=1. [default]`
which is not accepted by clingo. 

See the files: 
[visitor.txt](https://github.com/potassco/clingo/files/2388757/visitor.txt) 
and
[example.txt](https://github.com/potassco/clingo/files/2388759/example.txt)
and run
`gringo --text visitor.lp.`

rkaminsk:
@javier-romero commit 9b28871 introduces a backward breaking change. The ast will no longer print the [default] flag. Now it adds [override] whenever a constant definition has been given on the command line via `-c n=v`.

PR #624, opened 2026-04-10
Add Unicode and escape sequence support

### LINKED ISSUE #608: Only basic escapes are supported
```
inline std::string unquote(StringSpan str) {
    std::string res;
    bool slash = false;
    for (auto c : str) {
        if (slash) {
            switch (c) {
                case 'n': {
                    res.push_back('\n');
                    break;
                }
                case '\\': {
                    res.push_back('\\');
                    break;
                }
                case '"': {
                    res.push_back('"');
                    break;
                }
                default: {
                    assert(false);
                    break;
                }
            }
            slash = false;
        } else if (c == '\\') {
            slash = true;
        } else {
            res.push_back(c);
        }
    }
    return res;
}
```

Is there a specific why other escapes are not supported or could `unquote` and `quote` functions be extended with other escaped content, such as at least:
```
\t
\r
```

FrancoisLaferriere:
# Refs #123
## Changes
Added support for `\uXXXX`, `\t`, and `\r` escape sequences in strings.
- Lexer: Added `\t`, `\r`, and `\uXXXX` patterns to STRING and FLIT (f-strings)
- unquote(): Added handling for `\t`, `\r`, and `\uXXXX` escapes
- removed unused `quote()`
- PrintQuoted: Added `\r` escape handling
- Tests: Added tests for all new escape sequences

## Questions
* Should `\uXXXX` be converted to UTF-8 bytes in the internal representation (current behavior), or should it be preserved and printed back as `\uXXXX`? Currently:
   - `"caf\u00E9"` → internal UTF-8 → output `"café"`

* Currently, escape sequences in f-string literals are NOT processed, they pass through literally: `f"\n"` outputs `"\\n"`
Should we process escape sequences in f-string literals via `unquote()`, or keep the current behavior where they are passed through literally?


rkaminsk:
Regarding the questions:

1. Unicode escapes should be mapped to the UTF-8 sequences.
2. Escapes in f-strings should also be mapped. Maybe I forgot to do  this.

rkaminsk:
In the linked issue I was suggesting to terminate `\uX;` with a semicolon. This was intentional because the 4 digit representation does not cover all Unicode points. They range up to 10FFFF. Since we are free to come up with our own escape sequences, I would suggest to choose something that simply covers them all. If you don't like the terminating semicolon, I am also fine with another escape for arbitrary length ones, say `\u{X}`, for example, where X can be a hex digit in the range 0-10FFFF. These days emojis are too important to leave out. :wink:

lib/util/include/clingo/util/string.hh
```diff
@@ -1,43 +1,50 @@
...
+        return 10 + (c - 'A');
+    }
+    if (c >= 'a' && c <= 'f') {
+        return 10 + (c - 'a');
+    }
+    assert(false);
+    return 0;
+}
+
+inline auto encode_utf8(uint32_t cp, auto out) -> void {
```
rkaminsk, line 22: Add an assertion for the accepted input range or throw an exception.

lib/util/include/clingo/util/string.hh
```diff
@@ -1,43 +1,50 @@
...
-            }
-        }
+inline auto hex_val(char c) -> uint32_t {
+    if (c >= '0' && c <= '9') {
+        return c - '0';
+    }
+    if (c >= 'A' && c <= 'F') {
+        return 10 + (c - 'A');
+    }
+    if (c >= 'a' && c <= 'f') {
```
rkaminsk, line 16: I would typically turn the condition into an assertion or exception and simply return; or throw an exception at the end.

lib/util/include/clingo/util/string.hh
```diff
@@ -1,43 +1,50 @@
...
+    for (char h : hex) {
+        cp = (cp << 4) | hex_val(h);
+    }
+    encode_utf8(cp, out);
+}
+
 void unquote(std::string_view in, auto out, bool fstring = false) {
     auto escape = '\0';
-    for (auto c : in) {
+    for (size_t i = 0; i < in.size(); ++i) {
```
rkaminsk: Using indices here seems a bit clunky. If I cannot use ranges, I typically use iterators. I think the algorithm could be refactored to use iterators.

Since the function appends to an output stream, we could also use exceptions here and simply throw if parsing the string fails.

rkaminsk (changes requested):

rkaminsk:
@FrancoisLaferriere I refactored the unquoting function a bit. Since there is no need for a separate parse_unicode_escape function, I made it an implementation detail and made it partially parse string prefixes.

FrancoisLaferriere:
@rkaminsk I squashed the commits. It should be ready to merge now.

PR #629, opened 2026-04-24
Better handling of condition ids

rkaminsk:


lib/c-api/include/clingo/base.h
```diff
@@ -375,7 +375,11 @@ CLINGO_VISIBILITY_DEFAULT bool clingo_theory_base_element_condition(clingo_theor
                                                                     clingo_id_t element,
                                                                     clingo_literal_t const **condition, size_t *size);
 
-//! Get the id of the condition of the given theory element.
+//! Get a program literal for the condition of the given theory element.
+//!
+//! This returns zero for empty conditons (which is technically not a program
```
BenKaufmann: In general, I'm fine with this change. However, it promotes something that so far was an implementation detail somewhere deep down in lowest backend layer to the API level. That is, while some backends like `Clasp::Asp::LogicProgram` or `Potassco::AspifTextOutput` currently use `0` for the "empty" condition, the `Potassco::TheoryData` layer (i.e. the layer that this function here is using) does not assign any special semantics to id 0. In particular, it **does not guarantee** that 0 means "no"/"empty" condition. On this  layer, the only thing that is special about `0` is that it is stored in a more compact way (one bit instead of 4-byets) - but this is an non-observable implementation detail.

So, if we really want to have `0 = empty condition`, I should probably first adjust `libpotassco` accordingly.

rkaminsk: Maybe it would make sense to do this because the empty condition is the common case. <strike>It kind of makes sense because we do not have an inbuilt true literal in aspif, which probably lead to the whole mess in the first place.</strike> Condition ids are not represented in aspif but it makes sense to be able to refer to them via the API.

BenKaufmann (approved):

lib/c-api/include/clingo/base.h
BenKaufmann: @rkaminsk I clarified the contract in `clasp` and adjusted this PR accordingly. Please have a look at the [last commit](https://github.com/potassco/clingo/pull/629/changes/dd6536f6a5d2bfe9910ed35d3ac4e5acc4a6c0d8)

Note: I have not changed the behavior/guarantees of `Potassco::TheoryData` (i.e. there, `0` still has no fixed meaning) but this should be ok since the Clingo-C-API directly depends on `Clasp::LogicProgram` anyway. Hence, having the guarantee there should be enough.

PR #632, opened 2026-05-05
Fix index sequence

### LINKED ISSUE #630: Grounding drops rule depending on used syntax
Using the online preview of Clingo 6 (https://potassco.org/clingo-preview/), the program
```
true(V) :- set(V).
set(1) :- not true(2).
set(2) :- not true(1).
% --- variant 1
set(3) :- true(1..2).
% --- variant 2
%set(3) :- true(1;2).
% --- variant 3
%set(3) :- true(1).
%set(3) :- true(2).
```
produces the correct ground output (using "convert text").
If variant 2 or 3 are used instead of variant 1, the marked line is no longer contained in the ground output
```
set(1) :- not true(2).
set(2) :- not true(1).
true(1) :- set(1).
true(2) :- set(2).
set(3) :- true(1).
set(3) :- true(2).               <- this line is only present with variant 1
true(3) :- set(3).
```
When enumerating all models, as a consequence, the second model is incorrect.

With Clingo 5 (https://potassco.org/clingo/run/), the correct models are produced with
all program variants.

rkaminsk:


rkaminsk:
@copilot+claude-opus-4.5 Can you fix the following issues:
- There is a problem with cmake on the latest windows runner. Is the problem that the conda-forge cmake does not find the visual studio on this runner. I looked into the runner repo and it seems like vs 17 should be available.
- Can you refactor the index_sequence to not use a start index?

PR #650, opened 2026-08-13
implement #sort aggregate

MaxOstrowski:
This implements a sorting aggregate according to the discussion in https://github.com/potassco/clingo/discussions/469

Sorry for the large diff. It tries to cover all aspects of the new aggregate, including AST. There is a fast_track for domain input, otherwise the translation from the discussion is used.

As always I do not know what I'm doing, if anything is completely out of place or should have been done differently, please tell me.

Disclaimer: Most of the code changes are done with AI.


rkaminsk:
Can you also make your LLM describe what exactly it did and post it here?

MaxOstrowski:
Do you still want my LLM's summary (will probably be the same as the summary above).
Otherwise I'll be on vacation and will respond once I'm back.

rkaminsk:
As far as I can see, this implementation adds actual rules with conditional literals, when the sort aggregate is non-domain. I am not sure I like that. It would probably be better to do the translation internally.

MaxOstrowski:
https://github.com/MaxOstrowski/clingo/tree/sort_direct_translation
There is an alternative implementation that does the translation internally without relying on conditionals. Not sure if the semantic change in terms of incremental programs and convex aggregate is worth it. No need to react.

rkaminsk:
I am going to close this in favor of #666.

PR #654, opened 2026-08-15
`Windows.h` -> `windows.h` for cross-compiling with MinGW

ReubenJ:
Without this change, cross-compilation with MinGW fails because the file is actually `windows.h` and Linux file systems are case-sensitive. The build was successful on Windows because its file system is not case-sensitive.

### Checklist

**Please only open Pull Requests against the wip branch**

- [x] Have you followed the guidelines in our [CONTRIBUTING.md](/potassco/clingo/blob/master/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open Pull Requests for the same update/change?
- [ ] Have you made sure that all tests succeed?

### DIFF
--- a/libgringo/gringo/shared_mutex.hh
+++ b/libgringo/gringo/shared_mutex.hh
@@ -19,7 +19,7 @@
 #if GRINGO_PLATFORM_WINDOWS
 #define WIN32_LEAN_AND_MEAN
 #define NOMINMAX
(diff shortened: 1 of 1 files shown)


rkaminsk:
Thanks, i guess you can address this with a patch for now and we can delay another release.

PR #659, opened 2026-09-02
Make signed/unsigned conversions in the grounder explicit

GregoryGelfond:
This cleans up the implicit signed/unsigned conversions in clingo's own code under `lib/` and `app/`, so the whole tree builds cleanly under `-Wconversion -Wsign-conversion`.

**Why.** Each instance compiles just fine, but it makes the code difficult to both understand and reason about since each conversion requires a kind of mental reset, and it feels error prone in the long term. Making each conversion explicit removes that cost.

**What it does.** One uniform discipline, applied site by site rather than with a blanket cast:

- unsigned index/size types, and `std::cmp_*` for mixed-sign comparisons, where that's the right fix;
- at the real signedness boundaries — a `std::distance` used as a size, a literal known positive by an `assert`, a freshly minted id — an explicit `static_cast` with a one-line comment naming the invariant that makes it safe;
- the re2c-generated lexer include is wrapped in a diagnostic pragma, since the generated scanner reads a `char` buffer as `unsigned char` and isn't hand-maintained; the rest of that TU stays covered.

Representative before/after:

```cpp
// lib/ground/src/disjunction.cc — distance used as an index
-    return atoms_.find(sym) - atoms_.begin();
+    return static_cast<size_t>(atoms_.find(sym) - atoms_.begin());

// lib/control/src/solver.cc — literal known positive by the assert above it
     assert(lit > 0 && lit <= prg_lit_max);
-    return lit;
+    // lit > 0 (asserted above)
+    return static_cast<Potassco::Atom_t>(lit);
```

**Scope.** Behavior-preserving; every conversion here already happened implicitly and in range. It touches only `lib/` and `app/` (clingo's own code) — no public C API/ABI change, and it does **not** turn on any warning flag in the build; verifying the claim is just configuring with `-Wconversion -Wsign-conversion`. Two atom-id/literal casts are left with a `FIXME`: that pair is a genuine type conflation, and I'd rather fix it with distinct atom/literal types in a separate change than bury it in a cast here.

**Heads-up on where this goes.** This is deliberately just the mechanical cleanup. As a self-contained follow-up I'd like to give program atoms and literals small distinct value-types (a positive atom id vs. a signed literal, modelled on clingo's own `Symbol`), so the atom/literal conversions this PR makes *explicit* become named, total operations — and the two `FIXME`-marked id casts go away by construction rather than by comment. Roughly the shape, on one of the `std::abs` bridges in this diff:

```cpp
// this PR — explicit, but still bridges a literal to an atom by magnitude:
*is_fact = get_program(atoms).isFact(static_cast<clingo_atom_t>(std::abs(literal)));

// the follow-up — the conversion names what it means, and the raw form stays quarantined:
*is_fact = get_program(atoms).isFact(Literal::from_rep(literal).atom().index());
```

That's a separate PR on top of this one, not a dependency of it — I kept this change purely to the signedness pass so each stands on its own.

Tested with a clean `-Wconversion -Wsign-conversion` build and the full test suite.

### DIFF
--- a/app/main.cc
+++ b/app/main.cc
@@ -40,7 +40,7 @@ auto main(int argc, char *argv[]) -> int {
 #endif
     auto args = std::vector<clingo_string_t>{};
     try {
-        args.reserve(argc - 1);
+        args.reserve(static_cast<size_t>(argc - 1));
         // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic)
         std::ranges::transform(std::span{argv + 1, static_cast<size_t>(argc - 1)}, std::back_inserter(args),
                                [](auto const &x) { return clingo_string_t{x, strlen(x)}; });
--- a/lib/c-api/src/ast.cc
+++ b/lib/c-api/src/ast.cc
@@ -2946,7 +2946,7 @@ extern "C" auto clingo_ast_rewrite(clingo_ast_rewrite_context_t *context, clingo
         auto stm = convert<Stm>(statement);
         rewrite(context->ctx, stm, stms);
         ASTVec res{stms.size()};
-        int i = 0;
+        size_t i = 0;
         for (auto &stm : stms) {
             if (auto res_stm = unmap_params(*lib->store, context->param_unmap, stm); res_stm) {
                 stm = *std::move(res_stm);
--- a/lib/c-api/src/backend.cc
+++ b/lib/c-api/src/backend.cc
@@ -193,7 +193,8 @@ extern "C" auto clingo_backend_acyc_edge(clingo_backend_t *backend, int node_u, 
         if (backend == nullptr || (condition == nullptr && size > 0)) {
             return fail_arguments();
         }
-        get_program(backend).addAcycEdge(node_u, node_v, std::span{condition, size});
+        get_program(backend).addAcycEdge(static_cast<uint32_t>(node_u), static_cast<uint32_t>(node_v),
+                                         std::span{condition, size});
     }
     CLINGO_CATCH;
 }
@@ -204,7 +205,9 @@ extern "C" auto clingo_backend_add_atom(clingo_backend_t *backend, clingo_symbol
         if (backend == nullptr || atom == nullptr) {
             return fail_arguments();
         }
-        *atom = symbol != nullptr ? cpp_cast(backend)->add_atom(*cpp_cast(symbol)) : get_program(backend).newAtom();
+        // add_atom / newAtom yield fresh positive atom ids
+        *atom = symbol != nullptr ? static_cast<clingo_atom_t>(cpp_cast(backend)->add_atom(*cpp_cast(symbol)))
+                                  : get_program(backend).newAtom();
     }
     CLINGO_CATCH;
 }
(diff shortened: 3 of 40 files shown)


rkaminsk:
@GregoryGelfond, I agree that the code can be made more readable/robust in some places. However, I don't think that adding casts to unsigned serves this purpose here. They just add a lot of visual clutter without providing any safety guarantees.

When we do an index access, casting to unsigend serves no purpose except to silence a warning but does catch any errors. Some ideas what can be done:

If we want to make a signed value unsigend, we could define a helper function:

```
template <class T>
requires std::is_signed_v<T>
constexpr auto to_unsigned(T value) {
    assert(value >= 0);
    using U = std::make_unsigned_t<T>;
    return static_cast<U>(value);
}
```

Instead of using std::abs(lit), one could introduce a function:

```
constexpr atom_t lit_to_atom(lit_t lit) {
    assert(lit != 0);
    return static_cast<atom_t>(std::abs(lit));
}
```

Similarly, calls like `static_cast<size_t>(map_.find(sym) - map_.begin())` would better be done using a helper function:

```
template <class It>
constexpr auto forward_distance(It first, It last) {
    auto distance = std::distance(first, last);
    assert(distance >= 0);
    return to_unsigned(distance);
}
```

Strongly typed atom/literal types would make the code more robust. They typically come with a small performance penalty but that should not matter much for clingo.

GregoryGelfond:
Thanks — agreed. The casts add clutter without real safety, and typed atom/literal values are the better fix. I've already prototyped that: https://github.com/GregoryGelfond/clingo/compare/27c2615e...930c5305 — it adds `Atom` and `Literal` types and routes the atom↔literal crossings through typed accessors, so `std::abs(lit)` becomes `literal.atom()`. That covers your `lit_to_atom` directly; `to_unsigned` and `forward_distance` fit the remaining index/distance spots, and I'm glad to use those.

With your permission I'd rather pursue that than the casts here — happy to open it as its own PR or repoint this one, whichever you'd prefer. I'm still fairly new to contributing here, so I'll follow your lead on the structure.

rkaminsk:
This looks okay.

> https://github.com/GregoryGelfond/clingo/compare/27c2615e...930c5305

It actually makes the C++ code more robust. For the C API, I would not change anything because it is intended for higher level bindings anyway that provide their own safety mechanisms.

The from_rep function now could add an assertion to check that the literals/atoms are within range.

Just as a note, using typed enum classes would also have worked here. Conversion functions would have to be standalone functions then. However, using classes provides the most flexibility.

> With your permission I'd rather pursue that than the casts here — happy to open it as its own PR or repoint this one, whichever you'd prefer. I'm still fairly new to contributing here, so I'll follow your lead on the structure.

Sure go ahead if you have some time at hand. Contributions are welcome.

BenKaufmann:
@GregoryGelfond  @rkaminsk I find it a little weird that the strong types `Atom` and `Literal` create an invalid atom/literal by default. To me, this is a sign of a "leaky abstraction" - users of the classes would still have to check for rep is != 0 before use. For a proper abstraction, the default constructor would either have to be deleted or create a valid atom/literal. Additionally, the  `from_rep` would have to check for `rep != 0`. Of course, there are other ways to mitigate this problem but personally, I wouldn't bother.  At this (low) level of abstraction, I would probably either just keep the ints or use "typed" ints (aka enum classes like cf. `std::byte`).

rkaminsk:
> @GregoryGelfond @rkaminsk I find it a little weird that the strong types `Atom` and `Literal` create an invalid atom/literal by default. To me, this is a sign of a "leaky abstraction" - users of the classes would still have to check for rep is != 0 before use. For a proper abstraction, the default constructor would either have to be deleted or create a valid atom/literal. Additionally, the `from_rep` would have to check for `rep != 0`. Of course, there are other ways to mitigate this problem but personally, I wouldn't bother. At this (low) level of abstraction, I would probably either just keep the ints or use "typed" ints (aka enum classes like cf. `std::byte`).

Since this is low level code, there might be places where I used the invalid representation (even though I don't remember right now). Deleting the default constructor is still a good idea. It would show the relevant places and we could replace the invalid state by a more explicit invalid literal. I'd be fine with either a class or an enum based implementation. I don't see anything wrong with a class based variant as I think it improves readability.

GregoryGelfond:
Opened #661 with the Atom/Literal types on their own, off current wip-20. Per the discussion I've deleted the default constructor and use a named `Atom::invalid()` sentinel for the single placeholder site, so there's no invalid default state, and `from_rep` now asserts the value is non-zero and in range. The C API is unchanged. I kept the checked index/distance helpers (`to_unsigned`/`forward_distance`) out of it for now so it stays to the atom/literal typing; glad to add them as a follow-up. Closing this one in favour of #661.

Issue #583, opened 2025-11-13
clingo v6 hangs on reading


alalazo:
The `lp` files to reproduce are [asp.tar.gz](https://github.com/user-attachments/files/23519765/asp.tar.gz).

Using:
```console
$ clingo --version
clingo version 5.8.0
Address model: 64-bit

libclingo version 5.8.0
Configuration: with Python 3.13.5, without Lua

libclasp version 3.4.0 (libpotassco version 1.2.0)
Configuration: WITH_THREADS=1
Copyright (C) Benjamin Kaufmann

License: The MIT License <https://opensource.org/licenses/MIT>

$ clingo --verbose=3 --stats=2 --configuration=tweety --opt-strategy=usc,one,1 --heuristic=Domain --quiet=2,0,0 *.lp
clingo version 5.8.0
Reading from concretize.lp ...
------------------------------------------------------------------------------------------|
Call         : 1
Reading      : 7.900s
[ ... ]
OPTIMUM FOUND

Models       : 7
  Optimum    : yes
Optimization : 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 41 0 0 0 0 0 0 0 0 0 0 0 2 26 0 0 34 0 40 16 0 820 0
Calls        : 1
Time         : 24.690s (Solving: 12.91s 1st Model: 3.82s Unsat: 2.12s)
CPU Time     : 24.685s

[ ... ]
```

Using clingo v6:
```console
$ clingo --version
clingo version 6.0.0
Address model: 64-bit
libclasp version 4.0.0 (libpotassco version 2.0.0)
Configuration: WITH_THREADS=1
Copyright (C) Benjamin Kaufmann
License: The MIT License <https://opensource.org/licenses/MIT>

$ clingo --verbose=3 --stats=2 --configuration=tweety --opt-strategy=usc,one,1 --heuristic=Domain --quiet=2,0,0  --project-anonymous *.lp
clingo version 6.0.0
Reading from concretize.lp ...
------------------------------------------------------------------------------------------|
Call         : 1
Reading      : 
[ hangs here, Ctrl-C after a while ]
```

rkaminsk:
I'll have a look. This might be a grounding issue...

rkaminsk:
This seems to be some issue with projection, which I disabled in the run below using `--projection-mode=none`. I'll have to take a closer look later.

```
[...]

PR #660, opened 2026-09-03
Estimate empty recursive domains when ordering a rule body

GregoryGelfond:
## What this fixes

Issue #583 reports that grounding a particular program is very slow — a large Spack dependency-resolution encoding that grounds far more slowly than the size of the result would suggest.

The slowness is in how the grounder chooses the order to evaluate the literals in a rule body. It picks the order greedily, using a cheap estimate of how much work each literal will produce. For a **recursive** predicate — one still being computed in the current component — the grounder holds no atoms for it yet, so its domain size reads as **0**. Scored at size 0, such a literal looks trivially cheap, so the greedy order puts it first and defers the *selective* literals that would have bound its key arguments. The recursive literal then fans out over its whole relation before anything constrains it, and that intermediate blow-up is the cost.

## The change

When a rule body is linearized, take the **largest known domain size among the body's other literals** as the size to score such a not-yet-grounded recursive literal at, instead of 0. That keeps it from being mistaken for a trivially cheap literal, so the selective literals are picked first and the join stays small.

Nothing else changes: the greedy ordering algorithm, the semi-naive machinery, the queue, and the instantiators are untouched — only the size handed to a size-0 recursive literal differs. Every score stays non-negative, so the instantiation order remains valid and **the ground program is identical**. The ordering only affects *how much intermediate work* grounding does, never *what* it produces.

## Effect

Grounding the program from #583, before and after:

| | Rules | Atoms | Literal matches | Grounding time |
|---|---|---|---|---|
| before | 1,841,337 | 811,727 | 11,654,696 | 1.53 s |
| after  | 1,841,337 | 811,727 |  8,003,234 | 0.99 s |

*Generated with `clingo *.lp --stats --profile=detailed --project-anonymous --projection-mode=none --models=1 --quiet=2` on the #583 instance. Rules/Atoms are from `--stats`; "literal matches" (the join work) is summed from the `--profile=detailed` grounding profile; grounding time is the summed per-rule instantiation time, best of three runs. Platform: Apple M4 Pro (arm64), macOS 26.6.2, clang 23.1 + libc++, optimized (RelWithDebInfo) build.*

The ground program is byte-identical (Rules/Atoms/Bodies unchanged), so this is purely a grounding-speed change: about a third less join work and a third less grounding time. On a corpus of other programs the ground program was likewise unchanged and the answer sets identical.

Offered as a candidate fix for #583.

### DIFF
--- a/lib/c-api/src/control.cc
+++ b/lib/c-api/src/control.cc
@@ -83,7 +83,7 @@ extern "C" auto clingo_control_new(clingo_lib_t *lib, clingo_string_t const *arg
             clasp->startAsp(*slv_cfg, !opts.solver_options().single_shot);
         }
         auto slv = std::make_unique<CppClingo::Control::Solver>(*clasp, *slv_cfg, lib->log, *lib->store, lib->scripts,
-                                                                opts.rewrite_options(), opts.solver_options(), nullptr);
+                                                                opts.solver_options(), nullptr);
         opts.apply(*slv);
         *control = new clingo_control{lib, std::move(slv), std::move(slv_cfg), std::move(clasp)};
     }
--- a/lib/c-api/src/main.cc
+++ b/lib/c-api/src/main.cc
@@ -230,14 +230,8 @@ class ClingoApp : public Clasp::Cli::ClaspAppBase {
             if (opts_.mode() == AppMode::solve) {
                 clasp.startAsp(config(), false);
             }
-            auto slv = CppClingo::Control::Solver{clasp,
-                                                  config(),
-                                                  ctl_->lib->log,
-                                                  *ctl_->lib->store,
-                                                  ctl_->lib->scripts,
-                                                  opts_.rewrite_options(),
-                                                  opts_.solver_options(),
-                                                  stdout};
+            auto slv = CppClingo::Control::Solver{
+                clasp, config(), ctl_->lib->log, *ctl_->lib->store, ctl_->lib->scripts, opts_.solver_options(), stdout};
             opts_.apply(slv);
             // NOTE: member for createTextOutput
             ctl_->bind(&slv, &slv.config().clasp(), &slv.clasp_facade());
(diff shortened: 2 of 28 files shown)


rkaminsk:
This is interesting in so far that it is a simple tweak that seems to help here. It should only affect rules with multiple recursive literals as the literal that provides fresh instances per semi-naive iteration step is always ordered first. It is also a candidate for an option. I could imagine to choose the minium/maximum/average domain size as well. It generally makes sense not to use zero for the domain estimate as we do not exploit the structure of the domain otherwise. So even if the estimate is zero (for example if all literals in the rule body are recursive), we should set it to some value larger than zero.

GregoryGelfond:
Thanks — that matches what I saw on the "helps here" side, and I looked into the zero-estimate point specifically.

I tried flooring the all-recursive case (`body_max == 0` → score it at 1 instead of 0) and measured it across the kr-domains programs, including pure recursive self-joins like transitive reachability. As expected the ground program is unchanged — but so is the grounding work: literal matches and instances come out identical everywhere. The floor made no observable difference.

The reason is downstream in the scoring. A size-0 recursive literal's estimate flows into `Term::score`, and `TermFunction`/`TermTuple::do_score` already compute `root = std::max(1.0, std::pow(size / 2, 1.0 / len))`. That `std::max(1.0, …)` means an estimate of 0 and an estimate of 1 yield the same score — the "don't use zero" flooring effectively already happens there. And when every literal in the body is recursive they all receive the same estimate, so a small floor can't change their relative order in any case.

So an explicit `body_max` floor doesn't seem to buy anything, and I've left it out of this PR. The place with real signal to exploit is your min/max/average idea for when there *are* known-domain siblings — that moves the estimate to a value the scoring actually distinguishes. Happy to look at that, and the option surface, as a follow-up.

rkaminsk:
I refactored this a bit and made the estimate configurable. The option handling in the Solver is still a bit ugly. I'll still think about a better way...

PR #661, opened 2026-09-03
Add Atom and Literal types for program atom/literal ids

GregoryGelfond:
Program atoms and literals are carried as raw `prg_atom_t`/`prg_lit_t` integers, and the places that cross between them do it with hand-written sign arithmetic — `std::abs(lit)` to take the atom of a literal, a bare negation to flip a sign, a `static_cast` where the signedness changes. Nothing marks which integers are atoms and which are literals, so the crossings are easy to get subtly wrong and easy to misread.

This adds two small value types in `core/backend.hh`: **`Atom`** (a program atom, positive by construction; `index()`, `to_lit(sign)`) and **`Literal`** (a signed atom; `atom()`, `sign()`, `operator~`). The crossings go through these accessors — `std::abs(lit)` becomes `literal.atom().index()`, a sign flip becomes `~literal` — and the read side is fully typed. Both are trivially copyable and the same size as the underlying integer, converting to and from the raw representation only at the edges via `from_rep`/`to_rep`.

There is no invalid default state. The default constructors are deleted, so an `Atom` or `Literal` can only be formed from a real value — there is no zero-initialized instance a caller has to guard against. The one site that needs a placeholder (the atom-map entry inserted before its id is minted) uses a named `Atom::invalid()` factory that is overwritten with the real id on insertion and discarded otherwise, so it never reaches a read. `from_rep` asserts its argument is a non-zero id in range.

Scope: C++ internals only. The C API keeps its plain `uint32`/`int32` — it is intended for higher-level bindings that provide their own safety — converted at the `Potassco::*` seam. These are bespoke value-classes mirroring `Symbol`/`Number` rather than typed enums, so the accessors are real members. No behaviour change: ground programs and answer sets are identical, and the test suite passes (with new `roles.cc` unit tests).

This grew out of the discussion on #659, where the initial approach — unsigned casts to quiet conversion warnings — gave way to @rkaminsk's suggestion of strongly-typed atom/literal values. @BenKaufmann and @rkaminsk then pointed out that a default-constructed value was an invalid state callers had to account for; deleting the default constructor and the explicit `Atom::invalid()` above address that. The checked index/distance helpers (`to_unsigned`/`forward_distance`) sketched in that thread can follow as a separate step.

### DIFF
--- a/lib/ground/include/clingo/ground/base.hh
+++ b/lib/ground/include/clingo/ground/base.hh
@@ -245,7 +245,9 @@ class AtomBase : public BaseImpl<Symbol, AtomBase> {
 
     //! Add an atom to the base.
     template <class Gen> auto add(Symbol atom, StateAtom state, Gen &&gen) -> std::pair<MapAtom::iterator, AtomUpdate> {
-        auto [it, ins] = atoms_.try_emplace(atom, Atom{}, state);
+        // The placeholder is overwritten with the minted id below on insert, and
+        // discarded by try_emplace otherwise, so it never reaches a read.
+        auto [it, ins] = atoms_.try_emplace(atom, Atom::invalid(), state);
         if (ins) {
             it.value().id = Atom::from_rep(static_cast<prg_atom_t>(std::invoke(std::forward<Gen>(gen))));
             if (state != StateAtom::unknown) {
--- a/lib/ground/include/clingo/ground/base.hh
+++ b/lib/ground/include/clingo/ground/base.hh
@@ -3,6 +3,7 @@
 #include <clingo/ground/instantiator.hh>
 #include <clingo/ground/term.hh>
 
+#include <clingo/core/backend.hh>
 #include <clingo/core/symbol.hh>
 
 #include <clingo/util/index_sequence.hh>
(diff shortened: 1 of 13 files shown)


lib/ground/include/clingo/ground/base.hh
```diff
@@ -41,8 +41,8 @@ enum class StateAtom : uint64_t {
 
 //! Capture the state of an atom.
 struct AtomInfo {
-    uint64_t id : 62;    //!< A unique id among all atoms.
-    StateAtom state : 2; //!< The atom state.
+    Atom id;         //!< A unique id among all atoms.
+    StateAtom state; //!< The atom state.
 };
```
rkaminsk, line 46: This is really bad and would have double storage requirements.

rkaminsk:
Unfortunately, this cannot be merged in the current state. Clingo uses size_t to represent atoms, which is usually a 64bit integer. 32bit integers are mostly used by the backend because clasp uses 32bit integers to represent atoms. This PR now jumps between size_t and uint32_t more than once.

If we really want to get this merged, then all relevant places where size_t is used should use the Atom class as well.

PR #666, opened 2026-09-11
add #sort aggregates

rkaminsk:
Add #sort aggregates across parsing, rewriting, grounding, output, and public APIs, with broad test coverage.

## Changes:

- Implements #sort parsing, AST representation, grounding, and backend translation.
- Exposes the feature through C, C++, and Python APIs.
- Adds semantic, rewrite, safety, output, and integration tests.

## Example

```lp
d(1..3).
p(X) :- X = #sort { Y: d(Y) }.
```


rkaminsk:
@copilot

This is an LLM generated PR that needs some further refinements:

An `(X,Y) = #sort { ... }` aggregate has been added to sort atoms. For example:

The program
```
p(1..3).
s(X,Y) :- (X,Y) = #sort {Z:p(Z)}.
```
has solution
```
p(1) p(2) p(3) s(1,2) s(2,3)
```

The PR also accepts default negated sort aggregates but there is not a single test for them. I see two options to proceed here:

1. either already disallow negated sort aggregates in the parser
2. properly handle the negation

I  am actually in favor of option 1. because it is simpler and I do not see a need for negated sort aggregates.

The PR requires the term on the left hand side of the sort aggregate to be a pair of terms. I think this just makes things complicated in the implementation. We can just require a single term and use clingo term symbol matching to bind variables in the term. If the term is neither a tuple nor a variable it won't match anything. This is something that can be checked during simplification.

Clingo adds a lot of tests for unpooling, simplification, etc. The PR adds too few tests. Can you add more as is done for the other body literal types in clingo.

